PR #58, opened 2021-08-30
Add catalog hovers

DY-JS:
[DEMO LINK](https://DY-JS.github.io/layout_catalog_hovers/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "catalog-page",
+  "name": "catalog-page-grid",
   "version": "1.0.0",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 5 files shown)


vanyalarin:
Работа замечательная. Ты молодец)

Ihor-Karpyn (changes requested):
- `Smoothly (during 300ms)` increase the card by 20 percent on hover (neighboring cards **should not be** affected)
- `Smoothly (during 300ms)` change card title text color to `#34568b`
  - `:hover` should be applied only to the card, do not add it to the title directly
- `Smoothly (during 300ms)` change navigation link text color to `#00acdc`

Ihor-Karpyn (approved):
Great

PR #64, opened 2021-09-04
Catalog hovers

WorkAccCom:
- [DEMO LINK](https://workacccom.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,8 +14,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-  }
 
+    &:hover {
+      transform: scale(1.2);
+      @extend %hover-duration;
+      color: #34568b;
+    }
+      
+  }
+  
   &__card:not(:nth-child(4n)) {
     margin-right: 48px;
   }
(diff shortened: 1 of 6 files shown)


src/styles/_catalog.scss
```diff
@@ -14,8 +14,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
```
aliceblunt8, line 16: add transition just on card class and remove it from hover styles, that way your card will smoothly transformed in both ways (on hover and after hover)

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    &:hover {
+      transition-duration: 300ms;
```
aliceblunt8, line 42: same here, add transition to just class, remove transition from hover

src/styles/utiles/_extends.scss
```diff
@@ -0,0 +1,3 @@
+%hover-duration {
+  transition-duration: $transition-duration;
```
aliceblunt8, line 2: instead just transition-duration, use transition short-hand property and add property which transitions applies to

for example `transition: color 300ms ease-out` or `transition: transform 300ms ease-out`

aliceblunt8 (changes requested):

PR #115, opened 2021-10-18
Hover

ghost:
- [DEMO LINK](https://junglq.github.io/layout_catalog_hovers/)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,11 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all 0.3s ease 0s;
```
AlinaMatuschak, line 17: avoid transition `all`. You can replace it with a style name that will change smooth

PR #116, opened 2021-10-18
hover

mariia-bila:
[DEMO LINK](https://mariia-bila.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,14 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 300ms;
+
+    &:hover {
+      color: #34568b;
+      transform: scale(1.2);
+      transition-duration: 300ms;
+
+    }
   }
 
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,14 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 300ms;
```
dmtrhrytsak, line 17: Забула добавити плавну анімацію для transform:)

```suggestion
    transition: color 300ms, transform 300ms;
```

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 300ms;
```
ihor-jpeg, line 40: might be a good idea to create a variable, as usually the transition duration is the same across multiple elements

ihor-jpeg (approved):
Good stuff!

PR #117, opened 2021-10-18
Develop

Val-space:
[DEMO LINK](https://Val-space.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,7 +14,7 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 0.3s, color 0.3s;
```
ihor-jpeg, line 17: it might be a good idea to create a variable for the duration

ihor-jpeg (approved):
Good stuff!

PR #123, opened 2021-10-20
Develop

JamesSunderlanddd:
[DEMO LINK](https://JamesSunderlanddd.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -5,6 +5,8 @@
   max-width: 944px;
   margin: 0 auto;
 
+  cursor: pointer;
+
   &__card {
     display: flex;
     justify-content: center;
@@ -23,7 +25,6 @@
 
   &__card:hover {
     transform: scale(1.2);
-    cursor: pointer;
     color: #34568b;
   }
 }


src/styles/_catalog.scss
```diff
@@ -14,9 +14,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms;
```
dariia-romanova, line 17: Now your transition is applied only to transform property, color also have to change with duration of 300ms.

dariia-romanova (approved):

src/styles/_catalog.scss
```diff
@@ -14,9 +14,16 @@
...
+    transition: color 300ms, transform 300ms;
   }
 
   &__card:not(:nth-child(4n)) {
     margin-right: 48px;
   }
+
+  &__card:hover {
+    transform: scale(1.2);
+    cursor: pointer;
```
ihor-jpeg, line 26: we want the pointer cursor even if the hover effect is removed

ihor-jpeg (approved):
Good work. Let's fix that one little thing

PR #125, opened 2021-10-21
completed catalog hovers

i-stanko:


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 0.3s, transform 0.3s;
```
alex-kostomarov, line 17: perhaps the value in milliseconds would be easier for perception

PR #317, opened 2022-01-30
hover

maksym-yerokhin:
- [DEMO LINK](https://maksym-erokhin.github.io/layout_catalog_hovers/)


DarkMistyRoom (changes requested):
Good job, but use a variable. You should do it according to the task and it is useful because you have the same value in different places

DarkMistyRoom (approved):

PR #321, opened 2022-02-04
catalog_hover_ver

yahyafullstack01:
- [DEMO LINK](yahyafullstack01.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+  
+    &:hover {
+      color: #34568b;
+      border-radius: 20%;
+      transition: 0.3s;
+    }
   }
 
   &__card:not(:nth-child(4n)) {
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    
+    &:hover {
+      color: #00acdc;
+      transition: 500ms;
+    }
   }
 }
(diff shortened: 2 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+  
+    &:hover {
+      color: #34568b;
+      border-radius: 20%;
+      transition: 0.3s;
```
varseniuk, line 21:
```suggestion
      transition: 0.3s;
```
to work both ways, transition property should be applied not on :hover but on element itself.
Also, tou should always specify the `transition-property`
https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property

varseniuk (changes requested):

src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-property: 0.3s;
```
varseniuk, line 17: Why did you leave only transition-property? Transition-property should only specify the property, for wich the transition will be applied (not the duration)
It also should be transition-duration, because now there's no transition at all.
There is a shorthand like `transition: color 0.3s`, and it also may be written like
```
transition-property: color;
transition-duration: 0.3s;
```

https://user-images.githubusercontent.com/66570142/152564129-1395315b-06d2-49b1-863d-dee3ae57c9c3.mov

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    
+    &:hover {
+      color: #00acdc;
+      transition: 500ms;
```
varseniuk, line 43: specify property here as well

varseniuk (changes requested):

danylo-onishchuk (changes requested):
something wrong with demo link

danylo-onishchuk (changes requested):
increase the card by 20 percent on hover 
 
 use scale

PR #362, opened 2022-02-27
catalog_hovers was created

maxxnikiforov:
- [DEMO LINK](https://maxxnikiforov.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -4,6 +4,9 @@
 
   max-width: 944px;
   margin: 0 auto;
+  &__text {
+    transition-duration: 300ms;
+  }
 
   &__card {
     display: flex;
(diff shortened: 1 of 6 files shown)


Olexandr6 (changes requested):
When I click on the DEMO link, there is the 404 error, fix this

src/styles/_header.scss
```diff
@@ -37,5 +37,9 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
Olexandr6, line 40: Create a variable for this value, you use it in the several places

src/styles/_catalog.scss
```diff
@@ -14,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $duration-transform;
```
Olexandr6, line 20:
```suggestion
    transition: transform $transition-duration;
```

src/styles/_header.scss
Olexandr6, line 40:
```suggestion
    transition-duration: $duration-transform;
```

PR #384, opened 2022-03-15
Develop

SaperZP:
[DEMO LINK](https://saperzp.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms;
+  }
+
+  &__card:hover {
+    transform: scale(1.2);
+    color: #34568b;
   }
 
   &__card:not(:nth-child(4n)) {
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms;
```
AmbeMO, line 17:
```suggestion
    transition: 300ms;
```
```suggestion
    transition: transform 300ms, color 300ms;
```

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 300ms;
```
AmbeMO, line 40: same here, add property to transition

AmbeMO, line 40: same here, add property to transition

AmbeMO (changes requested):
you should add property to transition

src/styles/_catalog.scss
SaperZP: thanks you are totally right

PR #387, opened 2022-03-15
looks ready

yanaost:
https://yanaost.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
+    }
   }
 
   &__card:not(:nth-child(4n)) {
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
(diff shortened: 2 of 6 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
AmbeMO, line 17: you should add which style you need to transition
```suggestion
    transition: color 300ms, transform 300ms;
```

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
AmbeMO, line 40: same here

AmbeMO (changes requested):
You need to add property in transition

src/styles/_header.scss
AlinaMatuschak: add transition property )

AlinaMatuschak (changes requested):

src/styles/_header.scss
AlinaMatuschak: Always add a transition for styles that you want smoothly change

AlinaMatuschak, line 40: Always add a transition for styles that you want smoothly change

AlinaMatuschak (changes requested):

PR #399, opened 2022-03-24
Hovers were added

OlhaArama:
- [DEMO LINK](https://OlhaArama.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -37,5 +37,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
+    transition-timing-function: linear;
+  }
+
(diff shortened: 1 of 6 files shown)


src/styles/_header.scss
```diff
@@ -37,5 +37,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
+    transition-timing-function: linear;
```
ArturErmolenko, line 41: When you use transform add transition-property
https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property

ArturErmolenko (changes requested):

ArturErmolenko (approved):

PR #408, opened 2022-03-28
initial commit

amitpatiljc:
- [DEMO LINK](https://amitpatiljc.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    
+    transition-duration: 300ms;
+
+    &:hover {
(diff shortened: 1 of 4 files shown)


AndreiSuc (approved):
Good job

src/styles/_catalog.scss
```diff
@@ -14,6 +14,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+
+    transition-duration: 300ms;
```
ihor-jpeg, line 18: you should specify which properties you apply the transition to. in this case `all` is used which is not quite ok
```suggestion
    transition: transform 300ms, color 300ms;
```

ihor-jpeg (approved):
Well done!

PR #409, opened 2022-03-28
Develop

MabuzaM:
[DEMO LINK](https://MabuzaM.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -25,5 +25,6 @@
   &__card:hover {
     transform: scale(120%);
     color: #34568b;
+    cursor: pointer;
   }
 }
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,9 +14,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all ease;
+    transition-duration: 300ms;
   }
 
   &__card:not(:nth-child(4n)) {
     margin-right: 48px;
+  }
+
+  &__card:hover {
+    transform: scale(120%);
+    color: #34568b;
   }
 }
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -16,7 +16,6 @@
 .nav {
   &__list {
     display: flex;
-    margin: 0;
     padding: 0;
   }
 
@@ -41,7 +40,7 @@
 
   &__link:hover {
     transition: all ease;
-    transition-duration: 300ms;
+    transition-duration: $t-duration;
     color: #00acdc;
   }
 }
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -38,4 +38,10 @@
     font-size: 12px;
     font-weight: 500;
   }
+
+  &__link:hover {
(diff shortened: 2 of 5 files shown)


src/styles/_header.scss
```diff
@@ -37,5 +36,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color ease;
+  }
+
+  &__link:hover {
+    transition-duration: $t-duration;
```
59MrRobot, line 43: I recommend putting this line in the class and not the hover state so that it also transitions out smoothly when you "unhover". Same way you did it in the catalog class.

59MrRobot (changes requested):
Good job, just one suggestion.

src/styles/_header.scss
MabuzaM: Okay noted thanx done.

59MrRobot (approved):
Great job

src/styles/_header.scss
ihor-jpeg: yeah, but why do you also need it here tho?

src/styles/_header.scss
```diff
@@ -37,5 +36,12 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color ease;
+    transition-duration: $t-duration;
```
ihor-jpeg, line 40: you are already using shorthand prop. squeeze the duration in there also

src/styles/main.scss
```diff
@@ -1,14 +1,17 @@
...
+body,
+ul {
   margin: 0;
 }
 
 * {
   box-sizing: border-box;
 }
 
+$t-duration: 300ms;
```
ihor-jpeg, line 14: define it in :root or html

src/styles/main.scss
```diff
@@ -1,14 +1,17 @@
...
+
 @import "_header.scss";
 @import "_catalog.scss";
```
ihor-jpeg, line 17: why use `_`?

ihor-jpeg (changes requested):

ihor-jpeg (commented):
it does not seem like you have transform in your code but it exists on the deployed page. weird

src/styles/main.scss
MabuzaM: Done. Thank you

src/styles/_header.scss
MabuzaM: Fixed

src/styles/_header.scss
MabuzaM: Fixed

src/styles/_catalog.scss
```diff
@@ -14,9 +15,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all ease $t-duration;
```
meforoma, line 18: avoid using `all` - better specify particular ones
```suggestion
    transition: transform ease $t-duration;
```

meforoma (changes requested):

meforoma (approved):

PR #410, opened 2022-03-28
added smooth transitions and updated readme

LukaszGryziewicz:
- [DEMO LINK](https://LukaszGryziewicz.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,7 +14,7 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition-duration: 300ms;
+    transition-duration: $standard-transition-duration;
 
     &:hover {
       transform: scale(1.2);
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -37,7 +37,7 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
-    transition-duration: 300ms;
+    transition-duration: $standard-transition-duration;
 
     &:hover {
       color: #00acdc;
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -10,5 +10,7 @@ body {
   box-sizing: border-box;
 }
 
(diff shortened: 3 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
rHashTag, line 17: [SASS] - use variables for the main values so that you'll be able to reuse them and give them descriptive names. But don't overuse them, don't create variable for the value that's used just once.

`Transistion-duration` of `300ms` is also used in `src/styles/_header.scss`.

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
rHashTag, line 40: [SASS] - use variables for the main values so that you'll be able to reuse them and give them descriptive names. But don't overuse them, don't create variable for the value that's used just once.

`Transistion-duration` of `300ms` is also used in `src/styles/_catalog.scss`.

rHashTag (changes requested):
Well done, the code looks great!

There's really only one change I've suggested, have a look.

src/styles/_header.scss
LukaszGryziewicz: Thank you, that's a valid point.

LukaszGryziewicz:
Added variable for transition duration in the main.scss, felt like adding a separate file for it would be an overkill.

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $standard-transition-duration;
```
ihor-jpeg, line 40: specify the transition and which properties you apply it to

ihor-jpeg (approved):

PR #415, opened 2022-03-30
added hovers on card

iLebo-stack:
https://iLebo-stack.github.io/layout_catalog_hovers/


src/styles/Blocks/card.scss
```diff
@@ -0,0 +1,90 @@
...
+
+  &__title {
+    @include fonts($size: 12px, $line: 18px);
+    color: $secondary-color;
+    margin-bottom: 4px;
+    margin-top: $removing-default-margins;
+    transition: color 300ms;
+
+    &:hover {
+      color: #34568b;
```
andreisucman, line 29: This hover effect and color transition should be applied to the card, not the title. You can check this requirement on task page - point #3. At the moment you apply the hover effect to the title.

This is important because the title is much smaller than a card, so the transition won't kick in until the user hovers exactly over it, which is not ideal.

andreisucman (changes requested):
Hi, your code is working, but you missed an important detail. Please see my comment below.

Also, I noticed that you imported a lot of unnecessary code that makes checking your task harder. Consider importing only the necessary parts to make the checking easier for mentors./

Cheers.

ihor-jpeg (changes requested):
Well done! Take a look at Andrei's comment

src/styles/Blocks/card.scss
ihor-jpeg: I would argue that in this case, because he implemented the entire card, it doesn't make sense, at least to me, to highlight the text when hovering over the card. it might confuse the user and they'll think no matter where he clicks on the card, it will count as a click on the title

ihor-jpeg (approved):
whoops, wrong button

src/styles/Blocks/card.scss
andreisucman: If clicking on the title has a separate meaning or function then yes, it makes sense to highlight the title separately.

meforoma (approved):

PR #416, opened 2022-03-30
Catalog hovers task

proznerka:
- [DEMO LINK](https://proznerka.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,7 +14,7 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition-duration: 300ms;
+    transition: transform 300ms, color 300ms;
   }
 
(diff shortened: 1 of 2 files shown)


src/styles/_catalog.scss
```diff
@@ -14,9 +14,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
andreisucman, line 17: Ihor recommended specifying the name of the transition that you apply the duration on. For example `transition: color 300ms, transform 300ms;` At the moment you have` transition-property: all` (set by default because you didn't specify it); and `transition-duration: 300ms` applied to all; This is not ideal.

andreisucman (approved):
Good job, but please check my recommendation because Ihor is likely not to approve this code.

PR #418, opened 2022-03-31
solutionI

EnjSadman:
- [DEMO LINK](https://EnjSadman.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 0.3s;
+
+    &:hover {
+      transition-duration: 0.3s;
+      transform: scale(1.2, 1.2);
+    }
   }
 
   &__card:not(:nth-child(4n)) {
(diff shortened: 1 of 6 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 0.3s;
+
+    &:hover {
+      transition-duration: 0.3s;
```
ZUBOV-ILLIA, line 20: `transition-duration: 0.3s;`  already have this value in element, so its not necessary.

ZUBOV-ILLIA (approved):

src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 0.3s;
```
Anna-Tkachenko, line 17: You should explicitly specify properties which should change

src/styles/_catalog.scss
Anna-Tkachenko: Agree, fix it

Anna-Tkachenko (changes requested):
You didn't complete one point
image

src/styles/_catalog.scss
Anna-Tkachenko: You didn't specify changeable properties

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 👍

PR #421, opened 2022-03-31
made transforms

WojciechCzarnowski:
- [DEMO LINK](https://WojciechCzarnowski.github.io/layout_catalog_hovers/)

Tested myself, decided to pass

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,9 +14,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms;
+
+    &:hover {
+    transform: scale(1.2);
+    color: #34568b;
   }
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+    &__card:not(:nth-child(4n)) {
+      margin-right: 48px;
+    }
   }
 }
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -25,6 +25,11 @@
     color: #060b35;
     list-style: none;
     line-height: 60px;
+    transition: 300ms;
+
+    &:hover {
+      color: #00acdc;
+    }
   }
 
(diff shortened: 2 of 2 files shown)


src/styles/_catalog.scss
```diff
@@ -14,9 +14,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms;
```
proznerka, line 17: We met with reviews to better add something like this 

than default 'all'

proznerka (approved):
Good job :) 
Everything is ok. but I'm leaving a comment with what we met during ours reviews by mentors, take a look on it.

src/styles/_catalog.scss
meforoma: support @proznerka 
using `all` downside is it slows down load times

src/styles/_catalog.scss
```diff
@@ -14,9 +14,15 @@
...
+
+    &:hover {
+    transform: scale(1.2);
+    color: #34568b;
```
meforoma, line 21: you seem missing the card color and gaps between cards
not neat visually)
Image 2022-03-31 22 45 17

meforoma (changes requested):

meforoma (changes requested):
couldn't find what part of code to comment
please implement the color transitions
> Smoothly (during 300ms) change card title text color to #34568b
:hover should be applied only to the card, do not add it to the title directly

sorry, this one's fine - harder to notice

> Smoothly (during 300ms) change navigation link text color to #00acdc

this one doesn't seem to work

WojciechCzarnowski:
- [DEMO LINK](https://WojciechCzarnowski.github.io/layout_catalog_hovers/)

Did it twice by mistake

meforoma (approved):
😄👍  `fixed 'er up boss`

PR #424, opened 2022-04-01
catalog hover

PRO-GRAM-MER:
- [DEMO LINK](https://PRO-GRAM-MER.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,7 +14,7 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition-duration: 300ms;
+    transition: transform 0.3s;
 
     &:hover {
       transform: scale(1.2);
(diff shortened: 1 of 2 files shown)


meforoma (changes requested):
how about this requirement?
> Smoothly (during 300ms) change navigation link text color to #00acdc

src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
meforoma, line 17: you're implicitly using `any` transition - that slows down the page load times
better use shorthand
```suggestion
    transition: transform 0.3s, color 0.3s;
```
same for the header transition below

meforoma (changes requested):

meforoma (approved):

PR #426, opened 2022-04-02
catalog hovers

LTCebekhulu:
- [DEMO LINK](https://LTCebekhulu.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
+
+    &:hover {
+      color: #00acdc;
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
andreisucman, line 40: You probably need to add a transition name here too. You may think that it's obvious because you have just one transition here, but mentors recommend adding the name anyway. So` transition-duration: 300ms` should become `transition: color 300ms;` in my opinion.

andreisucman (approved):
Hi Lerato. Good job. I have just one suggestion that should help you get a faster approval.

ihor-jpeg (approved):
Well done! Take a look at Andrei's comment!

ihor-jpeg (approved):

PR #439, opened 2022-04-09
make some animation in catalog

uliton:
[DEMO LINK](https://uliton.github.io/layout_catalog_hovers/)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms;
+
+    &:hover {
+      color: #34568b;
+      transform: scale(1.2);
```
ZUBOV-ILLIA, line 21: option `transition` also must be for `color`

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 300ms;
+
+    &:hover {
+      color: #00addc;
```
ZUBOV-ILLIA, line 43: dont use `transition` for all, use only needed parametrs

ZUBOV-ILLIA (changes requested):

src/styles/_header.scss
uliton: ok)

src/styles/_catalog.scss
uliton: done

uliton: don`t understand at first time, but already change)

src/styles/_header.scss
uliton: i think, if we have only one change position, it can be used - "transition-duration" (as i made early)

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
Denchepornyuk, line 40: use transition for color instead of transition-duration

src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
```
Denchepornyuk, line 16: add cursor pointer to &__card

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #459, opened 2022-04-13
Solution

JustDan17:
[DEMO LINK](https://<your_account>.github.io/layout_catalog_hovers/)


src/styles/_catalog.scss
```diff
@@ -14,9 +14,17 @@
...
     margin-bottom: 48px;
+    cursor: pointer;
+    transition: 300ms;
   }
 
   &__card:not(:nth-child(4n)) {
     margin-right: 48px;
   }
+
+  &__card:hover {
```
rostyslav-pyliak, line 25: You can write this selector like &:hover. But, don't forgett about nesting.

src/styles/_catalog.scss
```diff
@@ -14,9 +14,17 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    cursor: pointer;
+    transition: 300ms;
```
rostyslav-pyliak, line 18: Add properties to which you apply transition

rostyslav-pyliak (changes requested):
Fix comments
And something wrong with DEMO LINK,  can't open it

src/styles/_catalog.scss
JustDan17: I can't do it, cause then it would trigger animation not on card but on the whole catalog.

src/styles/_catalog.scss
JustDan17: I can't do it here. I have more than one property for this transition time and it doesn't work otherwise.

src/styles/_catalog.scss
rostyslav-pyliak: Try to nest it in the " &__card " selector. It should work correctly

src/styles/_catalog.scss
rostyslav-pyliak: Try something like this: transition: transform 0.3s, color 0.3s;

JustDan17: Already figured. Just had to use coma

src/styles/_catalog.scss
JustDan17: Thx, it helped)

Denchepornyuk (approved):

PR #460, opened 2022-04-14
I have done)

Yaroslav-Kononenko:
- [DEMO LINK](https://Yaroslav-Kononenko.github.io/layout_catalog_hovers/)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms;
```
Denchepornyuk, line 17: use transition only for transform and color here

src/styles/_header.scss
```diff
@@ -37,5 +37,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: 300ms;
```
Denchepornyuk, line 41: use transition only for color here

Denchepornyuk (changes requested):

src/styles/_catalog.scss
Yaroslav-Kononenko: add transition-property as new line, but can`t add at one line due to it not work when more than one property (color, transform etc.) in line "transition"

PR #472, opened 2022-05-05
add transitions

nina-steigerwald:
[DEMO LINK](https://nina-steigerwald.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms linear;
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
+    }
   }
 
   &__card:not(:nth-child(4n)) {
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms linear;
```
aliceblunt8, line 17: specify properties you want to animate, + linear is default, no need to add it

```suggestion
    transition: color 300ms, transform 300ms;
```

aliceblunt8 (changes requested):

src/styles/main.scss
```diff
@@ -4,6 +4,7 @@ html {
 
 body {
   margin: 0;
+  transition: color 300ms, transform 300ms;
```
aliceblunt8, line 7: hmm why did you delete the transition from link and card? remove transition from body and add them back
but specify the properties you want to animate

aliceblunt8 (changes requested):

src/styles/main.scss
nina-steigerwald: My bad. Fixed

ArturErmolenko (approved):
Good job 👍

PR #477, opened 2022-05-09
add hover effects

Galyopa:
- [DEMO LINK](https://galyopa.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,7 +14,7 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition: all 300ms;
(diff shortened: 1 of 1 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all 300ms;
```
illiarerun, line 17: using `all` is a bit of bad practice. it'd be better to explicitly specify properties here

illiarerun (changes requested):
GJ 🔥 , one little fix left todo:

illiarerun (approved):

PR #482, opened 2022-05-12
Develop

vinnibea:
- [DEMO LINK](https://vinnibea.github.io/layout_catalog_hovers/)


src/styles/main.scss
```diff
@@ -10,5 +10,6 @@ body {
   box-sizing: border-box;
 }
 
+@import "./utils/placeholders.scss";
```
stanislav-topikha, line 13: You didn't add this file to commit

stanislav-topikha (changes requested):

src/styles/main.scss
vinnibea: Точняк!

src/styles/utils/placeholders.scss
```diff
@@ -0,0 +1,3 @@
+%basic-hover-animation {
+  transition: 300ms linear;
```
mykola-koval, line 2: `transition` is a shorthand property, use `transition-duration` if you only style the duration. And specify what properties you want to be affected by this rule.

mykola-koval (changes requested):

mykola-koval (approved):

PR #487, opened 2022-05-14
transformation was added

mflkvsk:
- [DEMO LINK](https://mflkvsk.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_utils.scss
+++ b/src/styles/_utils.scss
@@ -1,0 +1,13 @@
+$colors:
+  ".nav__link" #00acdc,
+  ".catalog__card" #34568b;
+
+@each $class, $color in $colors {
+  #{$class}:hover {
+    color: $color;
+  }
+}
+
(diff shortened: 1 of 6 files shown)


vsupruniuk (approved):

src/styles/_utils.scss
```diff
@@ -0,0 +1,13 @@
...
+  ".catalog__card" #34568b;
+
+@each $class, $color in $colors {
+  #{$class}:hover {
+    color: $color;
+  }
+}
+
+%duration {
+  transition: 300ms;
```
mykola-koval, line 12: transition is a shorthand property, use `transition-duration` if you only style the duration. And specify what properties you want to be affected by this rule.

mykola-koval (changes requested):

mykola-koval (approved):

PR #565, opened 2022-05-29
hover effects were added

misha-shevchuk:
- [DEMO LINK](https://misha-shevchuk.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -25,6 +25,12 @@
     color: #060b35;
     list-style: none;
     line-height: 60px;
+    transition: color 300ms;
+
+    &:hover {
+      color: #00acdc;
+      cursor: pointer;
+    }
   }
 
   &__item:not(:last-child) {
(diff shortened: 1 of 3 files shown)


src/styles/_header.scss
```diff
@@ -25,6 +25,12 @@
     color: #060b35;
     list-style: none;
     line-height: 60px;
+    transition: color 300ms;
+
+    &:hover {
+      color: #00acdc;
+      cursor: pointer;
```
Bohdan-Maksymiuk, line 32: isn't it default option for <a>?

Bohdan-Maksymiuk (approved):
Good job 👍

src/styles/_header.scss
misha-shevchuk: Hi, @Bohdan-Maksymiuk, what do you mean? If you mean style for cursor: pointer, then I added this to have pointer over all nav link block, not only over link's text. It looks prettier for me.

PR #583, opened 2022-06-09
hover v1

TanyaShevchenko:
- [DEMO LINK](https://TanyaShevchenko.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -15,7 +15,8 @@
     border-radius: 5px;
     margin-bottom: 48px;
     transition-duration: 300ms;
-
+    transition-property: width;
+ 
     &:hover {
       transform: scale(1.2);
       color: #34568b;
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -38,6 +38,7 @@
     font-size: 12px;
     font-weight: 500;
     transition-duration: 300ms;
+    transition-property: width;
 
     &:hover {
       color: #00acdc;
--- a/src/index.html
+++ b/src/index.html
@@ -24,7 +24,7 @@
       <nav class="nav">
         <ul class="nav__list">
           <li class="nav__item nav__item--is-active">
-            <a href="/" class="nav__link">Apple</a>
+            <a href="#" class="nav__link">Apple</a>
           </li>
           <li class="nav__item">
             <a href="/" class="nav__link">Samsung</a>


src/styles/_catalog.scss
```diff
@@ -14,9 +14,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
vkryvytskyy, line 17: Add transition-property

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
vkryvytskyy, line 40: add transition property

vkryvytskyy (changes requested):

src/styles/_catalog.scss
```diff
@@ -14,9 +14,17 @@
...
+    transition-property: width;
```
Bohdan-Maksymiuk, line 18: did you change width anywhere?

src/styles/_catalog.scss
```diff
@@ -14,9 +14,17 @@
...
+ 
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
```
Bohdan-Maksymiuk, line 22: These are the properties you change

Bohdan-Maksymiuk (changes requested):

src/styles/_catalog.scss
TanyaShevchenko: Frankly speaking, I don't understand why the 'transition-property' was requested if all "animations' were required to be smooth. According to the task, all changes under  ':hover' should be smooth with 300ms duration. So therefore what 'transition-property' should I use? Usage of 'transition-property: all' looks senseless because 'transition-property: all' is the default meaning

Could you kindly explain what exact  'transition-property' (transition-property: color or transition-property: transform or transition-property: all) do you ask me to use and why, because this change requirement isn't clear for me from the main task. Thank you in advance)

src/styles/_header.scss
```diff
@@ -37,5 +37,11 @@
...
+    transition-property: width;
```
ArtemTopolskyi, line 41: ты же анимируешь свойство color, почему здесь width?

src/styles/_catalog.scss
ArtemTopolskyi, line 18: color and transform. Write here those properties, that you animate

ArtemTopolskyi (changes requested):

src/styles/_catalog.scss
TanyaShevchenko: @Bohdan-Maksymiuk @vkryvytskyy Hello. Could you please give me some feedback?))))

Bohdan-Maksymiuk (changes requested):
Hi. Smooth hover doesn't work now - ping me in `fe_chat` or `fe_u22_****` and I could explain

src/styles/_catalog.scss
Bohdan-Maksymiuk: `all` is default and it works, but it consumes bigger amount of browser properties and it should affect the low performance machines. So better to indicate props  directly

PR #593, opened 2022-06-11
Develop

frontend112:


ihor-jpeg (changes requested):
Please deploy your project and make sure that the demo link is working and is present in the PR description
Screenshot 2022-06-15 at 01 16 45

meforoma (approved):
🥇 
+1 for NOT using `transition: all`
but particular transitions instead

PR #608, opened 2022-06-13
added hover effects to card and nav links

silchyn:
- [DEMO LINK](https://silchyn.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+
+    transition-duration: 300ms;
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+
+    transition-duration: 300ms;
+    transition-property: transform, color;
```
ivan-kadykalo, line 19: It would be better to use short syntax in the future
```suggestion
    transition: transform 300ms, color 300ms;
```

ivan-kadykalo (approved):
Good job💪

PR #624, opened 2022-06-15
Transition and transform styling of nav links and product cards applied

Ihor-Myslovskyi:
- [DEMO LINK](https://ihor-myslovskyi.github.io/layout_catalog_hovers/)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+
+    transition: $transition-duration ease;
```
ivan-kadykalo, line 18: You should write what parameter will use in this transition
For example: `transition: color 300ms`

ivan-kadykalo (changes requested):
Almost done

src/styles/_catalog.scss
Ihor-Myslovskyi: Hmmm, it worked in the way I wrote so I thought if it made my code shorter I could use it :)
But already checked your notes 👍

PR #646, opened 2022-06-16
hover effects were added

roman-romanenko:
[DEMO LINK](https://roman-romanenko.github.io/layout_catalog_hovers/)


src/styles/main.scss
```diff
@@ -1,14 +1,4 @@
...
-body {
-  margin: 0;
-}
-
-* {
-  box-sizing: border-box;
-}
+@import "./normalize";
 
 @import "_header.scss";
```
PavloMarkov, line 3: maybe it should be renamed - without "_" in the beginning (names of blocks don't have this "_"), and make 2 folders: "blocks" and "utils", to remove style-files there.

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 300ms;
```
PavloMarkov, line 40:
```suggestion
    transition: color 300ms;
```

PavloMarkov (approved):
good work 👍 
have some small comments.

src/styles/main.scss
PavloMarkov: comment was about underscore in style-names.

ivan-kadykalo: Don't need to use _ at the start of the name of your blocks. 
You can use that only with some additional files (variables, mixins etc) but not with a blocks

src/styles/_header.scss
ivan-kadykalo: Fix this comment

ivan-kadykalo (approved):
👍

PR #741, opened 2022-07-21
Develop

Andrii-Stiebalov:
[DEMO LINK](https://andrii-stiebalov.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 0.3s;
+
+    &:hover{
+      color: #34568b;
+      transform: scale(1.2);
+    }
   }
 
   &__card:not(:nth-child(4n)) {
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 0.3s;
+
+    &:hover{
+      color: #00acdc;
+    }
   }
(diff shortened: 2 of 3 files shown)


ShmelyovOleg (approved):

src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 0.3s;
```
AlinaMatuschak, line 17: Always add transition property. By default, transition property is all, and it's hard for the browser to control all properties

src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 0.3s;
```
AlinaMatuschak, line 40: here should add transition property that should change smoothly too )

AlinaMatuschak (changes requested):

src/styles/_catalog.scss
```diff
@@ -15,8 +15,9 @@
     border-radius: 5px;
     margin-bottom: 48px;
     transition-duration: 0.3s;
+    transition-property: color transform;
```
AlinaMatuschak, line 18: I suggest that it didn’t work correctly. Open dev tools and check (I can’t because you didn’t update the build and your demo link without the last changes)

But on dev tools something like this 👇
Screenshot 2022-07-22 at 20 24 16

AlinaMatuschak (changes requested):

AndriiRut (approved):
🔥

PR #815, opened 2022-08-09
transition

illia-sosnytskyi:
- [DEMO LINK](https://illia-sosnytskyi.github.io/layout_catalog_hovers/)


imaksweb (approved):
Good job! I can suggest you to add cursor pointer for a card hover state but it's completely optional

src/styles/blocks/catalog.scss
```diff
@@ -13,10 +12,19 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    transition: all 0.3s;
```
Anna-Tkachenko, line 15: don't use all with transition, it affects in bad way on website performance
specify only needed properties

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 🔥

PR #821, opened 2022-08-11
commited

devWizHub:
- [DEMO LINK](https://devWizHub.github.io/layout_catalog_hovers/)


amitpatiljc (changes requested):
You should just add and commit files in src folder.
You havn't pushed index.html 
All your cards are should scale, should not move sidewise.(read description)

devWizHub:
Screen Shot 2022-08-15 at 9 55 28 PM

the index.html file and the styles folder are within the src folder. its not clear why they are not uploading when deploying commiting ./src and deploying?  please advise.

devWizHub:
Use of scale argument for transform. Then followed point 13 - 18 for deploying the files into repository. thanks.

amitpatiljc (changes requested):
you are adding files with wrong command it not
git add ./src 
its 
git add src/. 
image

src/styles/_catalog.scss
```diff
@@ -14,9 +17,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
amitpatiljc, line 20:
```suggestion
        transition: transform 300ms, color 300ms;
```

amitpatiljc, line 20: you should specify which properties you apply the transition to. in this case all is used which is not quite ok

amitpatiljc (changes requested):

src/styles/_catalog.scss
devWizHub: I hope this solves, let's have a quick call in case I'm missing something. Thank you.

amitpatiljc (approved):
Good Job!

PR #824, opened 2022-08-13
Develop

diolna:
- [DEMO LINK](https://diolna.github.io/layout_catalog_hovers/)


src/styles/_catalog.scss
```diff
@@ -14,9 +14,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-  }
+    transition-duration: 300ms, 300ms;
+    transition-property: color, transform;
```
stanislav-topikha, line 18: value 300ms is the same for all transitions, use var
also consider using shorter syntax, a bit simpler to read for which prop you're applying transition
`transition: color 300ms, transform 300ms;`

src/styles/_header.scss
```diff
@@ -37,5 +37,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
stanislav-topikha, line 40: specify prop and use var

stanislav-topikha (changes requested):
Check the comments ⬇️

PR #826, opened 2022-08-14
develop

Blankevich:
https://Blankevich.github.io/layout_catalog_hovers/


src/styles/_catalog.scss
```diff
@@ -14,6 +14,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+
+    transition-property: color, transform;
+    transition-duration: 300ms;
```
stanislav-topikha, line 19: specify for which props transition durations applied
use var for time, it used more than once

stanislav-topikha (changes requested):
Check the comments ⬇️

src/styles/_catalog.scss
Blankevich: I don't quite understand what I need to do(((

alexandra-protyanova: it's fine but u can use more short version like:
transition: color 300ms, transform 300ms

alexandra-protyanova (approved):

PR #835, opened 2022-08-17
layout catalog hovers

constantweb:
[DEMO LINK](https://constantweb.github.io/layout_catalog_hovers/)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all 300ms linear;
```
marina-tilniak, line 17: it not good practice  use all, better use concrete properties

marina-tilniak (approved):

PR #836, opened 2022-08-17
Catalog with hover effect

MaxManis:
- [DEMO LINK](https://MaxManis.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -6,6 +6,7 @@
   margin: 0 auto;
 
   &__card {
+    position: relative;
     display: flex;
     justify-content: center;
     align-items: center;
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +15,14 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+
+    transform-origin: center;
+    transition-duration: 0.3s;
```
OleksiiBlezniuk, line 20: Better to define which exact properties will be animated, because if you not define it, it will be applied to every property that can be animated

src/styles/_catalog.scss
```diff
@@ -14,6 +15,14 @@
...
+
+    &:hover {
+      color: #34568b;
```
OleksiiBlezniuk, line 23: use scss variables for colors

OleksiiBlezniuk (changes requested):
GJ, left some comments

OleksiiBlezniuk (approved):
GJ!

PR #839, opened 2022-08-18
all hover effects from the task

TheVovchik:
- [DEMO LINK](https://TheVovchik.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,10 +14,11 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition: transform 300ms ease;
+    transition: all 300ms ease;
 
     &:hover {
       transform: scale(1.2, 1.2);
+      color: $transition-color;
     }
   }
 
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,11 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms ease;
+
+    &:hover {
+      transform: scale(1.2, 1.2);
+    }
   }
 
   &__card:not(:nth-child(4n)) {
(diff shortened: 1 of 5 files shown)


src/styles/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 300ms ease;
```
Anna-Borodaienko, line 40: easy is a default value, you don't need to set it

Anna-Borodaienko (commented):
GJ )))

src/styles/_catalog.scss
```diff
@@ -14,9 +14,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-  }
+    transition: transform $trans-duration ease-in-out,
+      color $trans-duration ease-in-out;
+
+    &:hover {
+      transform: scale(1.2, 1.2);
```
OleksiiBlezniuk, line 21:
```suggestion
      transform: scale(1.2);
```

src/styles/elements/_variables.scss
```diff
@@ -0,0 +1,3 @@
+$card-transition-color: #34568b;
+$nav-transition-color: #00acdc;
```
OleksiiBlezniuk, line 2: These colours can be used for other purposes too, so better to name them as it looks like (grey, blue, etc.) or as what they used for (text-main, text-secondary, etc.)

OleksiiBlezniuk (approved):
👍

src/styles/elements/_variables.scss
TheVovchik: now I start to name colors like you write, $page-color, $button-color or $color-grey, $color-light-red

src/styles/_catalog.scss
TheVovchik: agry

PR #854, opened 2022-08-21
Develop

Naveed-Shoukat:
- [DEMO LINK](https://naveed-shoukat.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -5,6 +5,8 @@
   max-width: 944px;
   margin: 0 auto;
 
+  perspective: 100px;
+
   &__card {
     display: flex;
     justify-content: center;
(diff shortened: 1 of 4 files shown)


kamal-tufekcic (approved):
index

src/styles/_catalog.scss
```diff
@@ -5,6 +5,8 @@
   max-width: 944px;
   margin: 0 auto;
 
+  perspective: 100px;
```
ihor-jpeg, line 8: but why?

ihor-jpeg (approved):

src/styles/_catalog.scss
Naveed-Shoukat: Agreed, there is no need of this property. 
Actually I was playing with this property and at the end I forget to remove this from the code.

Naveed-Shoukat: Agreed, there is no need of this property.
Actually I was playing with this property and at the end I forget to remove this from the code.

PR #856, opened 2022-08-22
Add hover effects for card and nav__links

Vladosik1337:
- [DEMO LINK](https://Vladosik1337.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/utils/_variables.scss
+++ b/src/styles/utils/_variables.scss
@@ -1,0 +1,3 @@
+$transition-duration: 300ms;
+$link-hover-color: #00acdc;
+$card-hover-text-color: #34568b;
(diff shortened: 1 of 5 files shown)


jojnajbude (approved):

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,3 @@
+$transition-duration: 300ms;
+$link-hover-color: #00acdc;
+$card-hover-text-color: #34568b;
```
OleksiiBlezniuk, line 3: These colours can be used for other purposes too, so better to name them as it looks like (grey, blue, etc.) or as what they used for (text-main, text-secondary, etc.)

OleksiiBlezniuk (approved):
GJ!

PR #863, opened 2022-08-22
Catalog hovers solution v01

anton-iskryk:
[DEMO LINK](https://anton-iskryk.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,9 +14,18 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all 300ms;
+
+    &:hover {
+      cursor: pointer;
+
+      transform: scale(1.2);
+      color: #34568b;
+    }
+
+    &:not(:nth-child(4n)) {
+      margin-right: 48px;
+    }
   }
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
-  }
 }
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -14,9 +14,18 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-  }
+    transition: all 300ms;
+
```
artem-nasadchuk, line 18: maybe better to define exactly which properties will be animated, not all?

artem-nasadchuk (approved):

src/styles/_catalog.scss
OleksiiBlezniuk: Better to define which exact properties will be animated, because if you not define it, it will be applied to every property that can be animated

src/styles/_catalog.scss
```diff
@@ -14,9 +14,18 @@
...
+    &:hover {
+      cursor: pointer;
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+      transform: scale(1.2);
+      color: #34568b;
```
OleksiiBlezniuk, line 23: Use variables for colors

OleksiiBlezniuk (changes requested):
GJ, but using `all` inside transition is a bad practice

OleksiiBlezniuk (approved):
Good job!

PR #879, opened 2022-08-23
hover transform styles was added

iPanda177:
- [DEMO LINK](https://iPanda177.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/utils/_normalizer.scss
+++ b/src/styles/utils/_normalizer.scss
@@ -1,0 +1,11 @@
+html {
+  font-family: Roboto, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+* {
+  box-sizing: border-box;
+}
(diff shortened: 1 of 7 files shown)


src/styles/utils/_normalizer.scss
```diff
@@ -0,0 +1,11 @@
...
+  font-family: Roboto, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+* {
+  box-sizing: border-box;
+}
```
iam-datsenko, line 11: I think it's better to write from * to body due to the hierarchy
And I consider Roboto have to be in " "

iam-datsenko (commented):
GJ, nothing to worry about! :)

MaksimusCRS (approved):
GJ

PR #896, opened 2022-08-25
Done

tatyana22:
- [DEMO LINK](https://tatyana22.github.io/layout_catalog_hovers/)


src/styles/_catalog.scss
```diff
@@ -14,9 +14,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
Yuriy-Komolov, line 17: I guess that you should make a variable for transition duration because you have that value more than once in your code.
And also would be good to add on hover effect cursor: pointer

Yuriy-Komolov (changes requested):

src/styles/_catalog.scss
tatyana22: ok.

Yuriy-Komolov (approved):

src/styles/blocks/_catalog.scss
```diff
@@ -14,9 +14,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $time;
```
Anna-Tkachenko, line 17: specify the properties for which you'll set transition

src/styles/blocks/_header.scss
```diff
@@ -37,5 +37,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $time;
```
Anna-Tkachenko, line 40: same here

Anna-Tkachenko (changes requested):

src/styles/blocks/_catalog.scss
tatyana22: ok

src/styles/blocks/_header.scss
tatyana22: ok

MaksimusCRS (approved):
GJ

PR #897, opened 2022-08-25
added hover and variables

volodymyrklymkiv:
- [DEMO LINK](https://volodymyrklymkiv.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/tools/_variables.scss
+++ b/src/styles/tools/_variables.scss
@@ -1,0 +1,4 @@
+$transition-duration: 0.3s;
+$transition-property: color;
+$link-color: #00acdc;
+$card-color: #34568b;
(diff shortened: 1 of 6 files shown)


Maksym845 (commented):
Good job!🚀

src/styles/tools/_variables.scss
```diff
@@ -0,0 +1,4 @@
+$transition-duration: 0.3s;
+$transition-property: color;
```
OleksiiBlezniuk, line 2: It's okay just to pass property every time, no need to save one particular prop in variable, but okay

I think better solution will be mixin that will take property as parameter, or placeholder. Or at least change name of this variable 🙂 It's definitely not clear what is property stored here

OleksiiBlezniuk (approved):
GJ!

PR #909, opened 2022-08-26
Develop

avkovlit:
[DEMO LINK](https://avkovlit.github.io/layout_catalog_hovers/)


src/styles/blocs/catalog.scss
```diff
@@ -0,0 +1,28 @@
...
+    justify-content: center;
+    align-items: center;
+    width: $card-width;
+    height: $card-height;
+    border: 1px solid $card-border-color;
+    border-radius: $card-border-radius;
+    margin-bottom: $card-margin;
+    transition: transform $transition-duration, color $transition-duration;
+
+    &:hover {
```
igor-1987, line 19: I guess it can be cursor: pointer on hover (optional).

src/styles/blocs/nav.scss
```diff
@@ -37,5 +22,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color $transition-duration linear;
```
igor-1987, line 25: This is optional thing, but may be you can play with type of timing functions. Linear is good, but with 300ms transition shows this effect too weak. You can try ease-in or ease-in-out. They are much more visible for such short duration.

igor-1987 (approved):
In general, nice job. My suggestions are optional for improving some tiny details.

src/styles/utils/_normalize.scss
```diff
@@ -0,0 +1,11 @@
+html {
+  font-family: Roboto, sans-serif;
```
Andr1yK, line 2: Mb, do we need to link this font family?)

Andr1yK (approved):

src/styles/utils/_normalize.scss
igor-1987: > Mb, do we need to link this font family?)

Like "font-family: $font-family"?

PR #912, opened 2022-08-26
catalog-hover

Arsen-Shpak:
[DEMO LINK](https://Arsen-Shpak.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/_catalog.scss
+++ b/src/styles/blocks/_catalog.scss
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: $transition-speed;
+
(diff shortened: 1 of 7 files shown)


vallerii:
Good job 👍

src/styles/blocks/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: $transition-speed;
```
OleksiiBlezniuk, line 17: Define properties that you want to animate, because now this transition is applied for every property that can be animated and we definitely don't want it

OleksiiBlezniuk (approved):

PR #929, opened 2022-08-28
the hover effect has been added

Ihor-1993:
[DEMO LINK](https://Ihor-1993.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,3 +1,5 @@
+$catalog-text-color: #34568b;
+
 .catalog {
   display: flex;
   flex-wrap: wrap;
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +16,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+
+    transition: 0.3s, transform, 0.3s;
```
Anna-Tkachenko, line 20:
```suggestion
    transition: color 0.3s, transform 0.3s;
```

Anna-Tkachenko (changes requested):
fix the correct setting transition duration for separate properties

PR #938, opened 2022-08-30
catalog hovers first attempt

andile-hlogwane:
- [DEMO LINK](https://andile-hlogwane.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,7 +1,6 @@
 .catalog {
   display: flex;
   flex-wrap: wrap;
-
   max-width: 944px;
   margin: 0 auto;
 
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +13,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all 300ms;
```
kamal-tufekcic, line 16: Bit risky in general, but okay here I suppose

kamal-tufekcic (approved):

src/styles/_catalog.scss
VinnieJ-2k20: If you transition just two properties - it's better to set them individually.
```suggestion
    transition: color 300ms, transform 300ms;
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #945, opened 2022-08-30
catalog-hover

IvanFront-end:
- [DEMO LINK](https://IvanFront-end.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_header.scss
+++ b/src/styles/_header.scss
@@ -1,3 +1,5 @@
+$Smoothly-duration: 0.3s;
+
 .header {
   display: flex;
   align-items: center;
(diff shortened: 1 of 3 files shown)


src/styles/_header.scss
```diff
@@ -1,3 +1,5 @@
+$Smoothly-duration: 0.3s;
```
stanislav-topikha, line 1: move it to the variables file, and use it in both files
also name it properly `transition-time` or smth

stanislav-topikha (changes requested):

SanyaBratashchuk (approved):
GJ

PR #947, opened 2022-08-30
added smooth transitioning

MeshackDev:
- [DEMO LINK](https://MeshackDev.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all 300ms ease-in;
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
+    }
   }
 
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all 300ms ease-in;
```
kamal-tufekcic, line 17: Don't know if it makes a difference for the task, but ease-in can be left out, also `transition: all` should be used carefully

kamal-tufekcic (approved):

src/styles/_catalog.scss
VinnieJ-2k20: Yes, use individual properties instead of `all`.
`all` should be used if you transition 4 or more rules at the same time.

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #979, opened 2022-09-08
task

NVlasenko:
- [DEMO LINK](https://NVlasenko.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,7 +1,6 @@
 .catalog {
   display: flex;
   flex-wrap: wrap;
-
   max-width: 944px;
   margin: 0 auto;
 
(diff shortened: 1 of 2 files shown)


src/styles/_catalog.scss
```diff
@@ -13,7 +12,13 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
+    transition-duration: 300ms;
```
alexandra-protyanova, line 15: set transition to specific properties not for everything

alexandra-protyanova (changes requested):

Virass (changes requested):
Not use transition for :hover, because animation work only for a hover effect, and after element, rollback to the initial state without animation.

Set transition-duration to catalog__card selector and set option, what effect needed animation.

AlisaVlasova (approved):

PR #983, opened 2022-09-11
catalog-hovers-task-done

YaTut1901:
- [DEMO LINK](https://YaTut1901.github.io/layout_catalog_hovers/)


marina-tilniak (changes requested):
add transition property and make sure to add transition style under general selector, not the one with :hover - this way transition will work smoothly both ways.

PR #987, opened 2022-09-14
commit

Serrggii:
https://Serrggii.github.io/layout_catalog_hovers


marina-tilniak (changes requested):
add transition property and make sure to add transition style under general selector, not the one with :hover - this way transition will work smoothly both ways.

marina-tilniak (approved):

PR #1056, opened 2022-10-06
Catalog hovers was added

MikhailKucherina:
- [DEMO LINK](https://MikhailKucherina.github.io/layout_catalog_hovers/)


Stepico (changes requested):
It seems pretty nice, but you did a lot of irrelevant work, you just had to follow that mockup you already have in your project while installing. It's well done, but I guess you should remake it. As you can see your device's name doesn't follow your style, because it's too hard for us now to change some properties of one element while we are interacting with another one

PR #1062, opened 2022-10-06
Catalog_Hovers

referee1988:
[DEMO LINK](https://referee1988.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 300ms, transform 300ms;
+
+    &:hover {
+      color: #34568b;
+      transform: scale(1.2);
+    }
   }
 
   &__card:not(:nth-child(4n)) {
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,6 +14,7 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    
     transition: color 300ms, transform 300ms;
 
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 300ms, transform 300ms;
```
Oleksii-Siviolov, line 17: Checklist, point 7.

src/styles/_header.scss
```diff
@@ -25,6 +25,11 @@
     color: #060b35;
     list-style: none;
     line-height: 60px;
+    transition: color 300ms;
```
Oleksii-Siviolov, line 28: Same here.

Oleksii-Siviolov (changes requested):
All works good in the demo link, the job looks good to me, but need to do some changes according to checklist, as I dont see variables.

src/styles/_catalog.scss
denys-danyliuk: 4 point in checklist actually, use variable

denys-danyliuk (changes requested):
fix comments

referee1988:
> fix comments

image

Point 1 of the Checklist contradicts point 4. Such a checklist does more harm than good!

denys-danyliuk:
> > fix comments
> 
> image
> 
> Point 1 of the Checklist contradicts point 4. Such a checklist does more harm than good!

this example is about totally different thing, and putting here a variable for this values would be not obvious

denys-danyliuk (approved):

PR #1063, opened 2022-10-07
Catalog Hovers Task is Solved

Stepico:
- [DEMO LINK](https://Stepico.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,9 +14,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-  }
+    transition: transform 0.3s, color 0.3s;
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+    &:hover {
+      cursor: pointer;
+      transform: scale(1.2);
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -14,9 +14,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-  }
+    transition: transform 0.3s, color 0.3s;
```
MikhailKucherina, line 17: u use transform duration 2 times, it will be better to make a vabiable.

MikhailKucherina (changes requested):
good job, fix one thing and it will be great!

src/styles/_catalog.scss
Stepico: yeah, maybe in this task it will be good, but I don't think like that about some huge projects, because someone might want to set them different values sometime. I guess it store in one place, so you can easily find this information and change it, why do we need to create another one "in the one place" as variables

MikhailKucherina (approved):

illiarerun (approved):

PR #1078, opened 2022-10-08
Make animation in catalog

bohdan-stepanets:


src/styles/_catalog.scss
```diff
@@ -4,6 +4,7 @@
 
   max-width: 944px;
   margin: 0 auto;
+  perspective: 100px;
```
sofiia-dovhan, line 7: You don't need perspective here in case when you use "transform: scale"
It used with "translate"

sofiia-dovhan (commented):
Good job! :fire:

sofiia-dovhan (approved):

Olexandr6 (approved):

PR #1087, opened 2022-10-10
DEMO LINK

Nataliia95254:
[DEMO LINK](https://Nataliia95254.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -19,4 +19,11 @@
   &__card:not(:nth-child(4n)) {
     margin-right: 48px;
   }
+
+  &__card:hover {
+    cursor: pointer;
+    transition-duration: 300ms;
+    transform: scale(1.2);
+    color: #34568b;
+  }
 }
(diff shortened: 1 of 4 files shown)


loralevitska (changes requested):
It's better to set the `transform` & `transition` property for the card class, so the Smoothly effect will be preserved when the mouse is placed on the card and removed from the card. Currently, this effect is preserved only when hovering over the card, and not in the reverse direction.

src/styles/_catalog.scss
```diff
@@ -14,9 +14,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
loralevitska, line 17: for future instead of &__card:hover

```suggestion
    transition-duration: 300ms;
    
    &:hover {
      cursor: pointer;
      color: #34568b;
      transform: scale(1.2);
    }
    
```

loralevitska (approved):
great :wink:

PR #1090, opened 2022-10-11
Develop

myroslava-kukhta:


src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
+  }
```
stanislav-topikha, line 18: Specify for which properties you're applying transition, move time to the var(you used it 2 times)

stanislav-topikha (changes requested):

src/styles/_catalog.scss
```diff
@@ -14,6 +14,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $transition-duration;
```
polosanya, line 17: Previous comment not fixed. Please specify for which property transition should be applied

polosanya (changes requested):
Almost done! Please pay attention to the comments

liliia-livandovskaya (approved):
🔥

PR #1096, opened 2022-10-13
Layout catalog hover

Liliya-Kalinichenko:
- [DEMO LINK](https://Liliya-Kalinichenko.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $transition-time;
```
Bohdan-Maksymiuk, line 20: You should specify transition properties directly

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $transition-time;
```
Bohdan-Maksymiuk, line 41: You should specify transition properties directly

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1104, opened 2022-10-18
Created

IVxVI:
- [DEMO LINK](https://IVxVI.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms ease;
```
alexandra-protyanova, line 20: set transition to define properties and move duration to variable

alexandra-protyanova (changes requested):

src/styles/blocks/catalog.scss
IVxVI: It crashes when I try to create variable, only works when in exact file... So should it be like this?

DarkMistyRoom: That is because you should make `main.scss` file and import all your blocks and utils there

DarkMistyRoom: And if you have some questions, ask in fe_chat

DarkMistyRoom (changes requested):
Correct previous comment

src/styles/blocks/catalog.scss
IVxVI: Understood. It`s just import "variables" has to be on top)

alexandra-protyanova (approved):

PR #1108, opened 2022-10-19
commit message

diamat1997:
- [DEMO LINK](https://diamat1997.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
+
+    &:hover {
+      color: #34568b;
+      transform: scale(1.2);
+    }
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,7 +17,7 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition: 300ms;
+    transition: transform 300ms;
 
     &:hover {
       color: #34568b;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
alexandra-protyanova, line 41: set transition to define properties, not to `all`, and use variable for duration

alexandra-protyanova (changes requested):

src/styles/blocks/header.scss
diamat1997: Do you mean that you need to use `transition` on a `transition-duration`?Because I didn't quite understand what you meant.

src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms;
```
DarkMistyRoom, line 20: You should specify the property. 
for example: 
```suggestion
    transition: color 300ms;
```

DarkMistyRoom: But do it for all properties and use a variable for the duration

DarkMistyRoom: https://developer.mozilla.org/en-US/docs/Web/CSS/transition

DarkMistyRoom: And if you have a question, ask it in the fe_chat. It will be faster

DarkMistyRoom (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms;
+
+    &:hover {
+      color: #34568b;
```
alexandra-protyanova, line 23: I think, we should add transition for color as well

alexandra-protyanova (changes requested):

PR #1114, opened 2022-10-21
Animation training

Rio1391:
- [DEMO LINK](https://rio1391.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-
+    transition: all $trans-duration;
```
SanyaBratashchuk, line 20: do not use `all` write each property

src/styles/main.scss
```diff
@@ -1,3 +1,5 @@
+$trans-duration: 0.3s;
```
SanyaBratashchuk, line 1:
```suggestion
$transition-duration: 0.3s;
```

SanyaBratashchuk (changes requested):

src/styles/blocks/catalog.scss
Rio1391: done

src/styles/main.scss
Rio1391: done

SanyaBratashchuk (approved):

PR #1122, opened 2022-10-24
hovers were made

vlad-boikov:
- [DEMO LINK](https://vlad-boikov.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $transition-duration;
```
DarkMistyRoom, line 20: list all props, it is better for performance

DarkMistyRoom (changes requested):

liliia-livandovskaya (approved):

PR #1137, opened 2022-10-29
catalog hover

ruslan-kasinets:
[DEMO](https://ruslan-kasinets.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $transition-duration, color $transition-duration;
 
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
...
     margin-bottom: 48px;
+    transition: transform $transition-duration, color $transition-duration;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
```
polosanya, line 28: For colors it is also better to create variables

polosanya (changes requested):
Almost done! Please check my comments

loralevitska (approved):
good job!

PR #1148, opened 2022-11-02
Advance HTML and CSS Lesson 5 - Catalog Hovers

ysrckr:
- [DEMO LINK](https://ysrckr.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/utils/_variables.scss
+++ b/src/styles/utils/_variables.scss
@@ -1,0 +1,1 @@
+$transition-color: color 300ms ease-in-out;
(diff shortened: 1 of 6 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$transition-color: color 300ms ease-in-out;
```
mgrinko, line 1: You could use a mixin, but not the exact value with exact transition-property value

mgrinko (approved):

Vasya564 (approved):

PR #1153, opened 2022-11-04
adde hover

MSA1991:
- [DEMO LINK](https://msa1991.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all $duration;
+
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    cursor: pointer;
+    transition: all $duration;
```
ivanchuk-yurii, line 21: don't use `all`, set only for `transform` and `color`

ivanchuk-yurii (changes requested):

ivanchuk-yurii (approved):

PR #1172, opened 2022-11-08
added hovers

illiashcherbak:
- [DEMO LINK](https://illiashcherbak.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,17 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: 
+      transform $transform-duration $linear-transform,
+      color $transform-duration $linear-transform;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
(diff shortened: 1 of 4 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,17 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: 
+      transform $transform-duration $linear-transform,
+      color $transform-duration $linear-transform;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
```
vladimir-arkanov, line 27: I think, we will you this catalog for next topics and this parameter could be used again, so can be variables

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,17 @@
     border-radius: 5px;
     margin-bottom: 48px;
```
vladimir-arkanov, line 19: also, i think we will use it in some calculations in future

solaryasha (approved):
GJ!

PR #1175, opened 2022-11-09
catalog-hovers v1.0

Donovanoff:
[DEMO LINK](https://donovanoff.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 0.3s, transform 0.3s;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 0.3s, transform 0.3s;
```
vlad-demchuk, line 20: image

vlad-demchuk (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: $Smoothly;
```
marina-tilniak, line 20: it's a good practice to add transition popery, because now transition is applied for all properties

marina-tilniak (approved):

PR #1176, opened 2022-11-09
catalog-hover added

OlexandraKorol:
- [DEMO LINK](https://OlexandraKorol.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -11,6 +11,7 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
+    cursor: pointer;
 
     width: 200px;
     height: 300px;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -11,15 +11,23 @@
...
     box-sizing: border-box;
+    cursor: pointer;
 
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform 300ms ease, color 300ms ease;
```
Anastasia4sik, line 22: checklist p4. You can add transition duration as a variable since you use the same value for it more than once
Correct it in the whole doc

Anastasia4sik (approved):
Good job, just fix one thing.

src/styles/blocks/catalog.scss
OlexandraKorol: Thanks! Just fixed it!

src/styles/blocks/header.scss
```diff
@@ -38,5 +39,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: color 300ms ease;
```
solaryasha, line 43:
```suggestion
    transition: color $transitions-durations ease;
```

solaryasha (approved):
Good job!

PR #1177, opened 2022-11-09
added transform hover

Iksich:
[DEMO LINK](https://Iksich.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -26,7 +26,7 @@
     color: #060b35;
     list-style: none;
     line-height: 60px;
-    transition: 300ms;
+    transition: color 300ms;
 
     &:hover {
       color: #00acdc;


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms, color 300ms;
```
nurdanokcu, line 20:
```suggestion
    transition:
      transform 300ms, 
      color 300ms;
```
This suggestion is just to make code look better. If you want to add more properties later on this way it will look less confusing. But for this small project it is fine of course.

src/styles/blocks/header.scss
```diff
@@ -26,6 +26,11 @@
     color: #060b35;
     list-style: none;
     line-height: 60px;
+    transition: 300ms;
```
nurdanokcu, line 29:
```suggestion
    transition: color 300ms;
```
It is better to add color property here so we can know what the transition is for .

nurdanokcu (changes requested):
Everything looks good. I had 2 suggestions about transitions.

nurdanokcu (approved):

src/styles/main.scss
```diff
@@ -1,3 +1,3 @@
-@import "blocks/page";
+@import "blocks/page.scss";
 @import "blocks/header";
 @import "blocks/catalog";
```
meforoma, line 3: not really consistent imports
and i can't seem to find the `page.scss` file 🕵️

meforoma (approved):
all works 👍 
but please take a look and address the comment i've left for you

PR #1179, opened 2022-11-09
first commit hover catalog

vladimir-arkanov:
- [DEMO LINK](https://vladimir-arkanov.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -12,14 +12,21 @@
...
-    width: 200px;
-    height: 300px;
+    width: $card-width;
+    height: $card-height;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    margin-bottom: $margin-bottom-for-card;
+
+    transition: color 300ms linear, transform 300ms linear;
```
illiashcherbak, line 21: I guess these properties can be made variables, but that's okey.

src/styles/blocks/catalog.scss
```diff
@@ -12,14 +12,21 @@
...
 
     &:not(:nth-child(4n)) {
-      margin-right: 48px;
+      margin-right: $margin-right-for-card;
+    }
+
+    &:hover {
+      transform: scale(1.2, 1.2);
```
illiashcherbak, line 28: You can add only one value of scale 1.2 and it will work the same

src/styles/main.scss
```diff
@@ -1,3 +1,4 @@
+@import "./utils/variables";
 @import "blocks/page";
 @import "blocks/header";
 @import "blocks/catalog";
```
illiashcherbak, line 4: I think you need to change the path to the block folder

illiashcherbak (changes requested):
I reject your work only because I think you need to change the path to the block folder. Otherwise it's good job!

src/styles/blocks/catalog.scss
vladimir-arkanov: I put properties in variables, now looks nice

src/styles/blocks/catalog.scss
vladimir-arkanov: Also, deleted extra scale

src/styles/main.scss
vladimir-arkanov: Path was renewed

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,6 @@
+$card-width: 200px;
+$card-height: 300px;
+$margin-right-for-card: 48px;
+$margin-bottom-for-card: 48px;
+$properties-for-trantison: linear;
```
solaryasha, line 5: It's not properties for transition, it's transition function

solaryasha (approved):

PR #1204, opened 2022-11-10
transform and transition were made

Mykhailyk-Vadym:
- [DEMO LINK](https://Mykhailyk-Vadym.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,7 +18,7 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition: color, transform 300ms linear;
+    transition: color $time-of-transition linear, transform $time-of-transition linear;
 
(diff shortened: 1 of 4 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: color, transform 300ms linear;
```
4ebupel, line 21: Checklist p.4 use variables for the main values so that you'll be able to reuse them and give them descriptive names....
I think it's better to create a variable for the '300ms' value, fix in all cases (c)Ihor Karpyn)))

4ebupel (changes requested):
Good job, my code have the same problem, so I'm going to fix that too :)

4ebupel (approved):
Good job! Have a nice evening)

PR #1207, opened 2022-11-10
will see

Qsery93:
no tests for this exercise


marina-tilniak (changes requested):
we don't have test for this exercise, but where is demo link?
fix checklist point 1 and 4

Qsery93:
Changes added. But current demo-link doesnt accept the changes. npm run peloy has a lot of mistakes even after reinstall the exercise. This happended only on this exercise. Can u check it without demo ? Or i need to close the pull request and create new one ? That could fix it.

marina-tilniak (changes requested):
oh, ok Let's skip demo for this task, but you should fix checklist point 4 and move to variable transition duration

Qsery93:
Was added $smooth for transition duration.

tetiana-ivanchenko (approved):

PR #1220, opened 2022-11-11
Catalog hover

kattentionn:
- [DEMO LINK](https://kattentionn.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,4 +1,4 @@
-@import "blocks/_variables.scss";
+@import "utils/_variables.scss";
 @import "blocks/page";
 @import "blocks/header";
 @import "blocks/catalog";
--- a/src/styles/utils/_variables.scss
+++ b/src/styles/utils/_variables.scss
@@ -1,0 +1,1 @@
+$animation-time: 300ms;


src/styles/blocks/_variables.scss
```diff
@@ -0,0 +1 @@
+$animation-time: 300ms;
```
YakovlewDan, line 1: it's ok but remove this file from 'blocks' folder

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,14 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $animation-time;
```
YakovlewDan, line 20: It would be best if you explicitly named all properties that will change on transition, like:
```suggestion
    transition: property1 $animation-time, 
    	property2 $animation-time ;
```

YakovlewDan (changes requested):
Looks ok, but a few changes must be done IMO.

src/styles/blocks/_variables.scss
kattentionn: ok, i will move. thank you

src/styles/blocks/catalog.scss
kattentionn: Thank you, but there are only 2 changes to the animation; without them, the code will be shorter. But maybe I'm not right:)

valerii-derkach, line 20: yes in our case it is only two properties and it is shorter, but for the future it is better to name exact properties that will change on transition, to avoid mistakes, when the behaviour of our code is unexpectred

valerii-derkach (approved):

valerii-derkach (approved):

src/styles/blocks/catalog.scss
kattentionn: Okay, thank you! I will know in the future)

PR #1222, opened 2022-11-11
initial commit

44shaggy:
- [DEMO LINK](https://44shaggy.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: 0.3s;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: 0.3s;
```
qoa11a, line 21: Let's apply transition only to the required properties rather than to everything.

qoa11a (changes requested):

src/styles/blocks/catalog.scss
44shaggy: got it

amitpatiljc (approved):
Good Job!

PR #1228, opened 2022-11-11
Created catalog_hovers

nazarkharkevych:
https://nazarkharkevych.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/utils/_variables.scss
+++ b/src/styles/utils/_variables.scss
@@ -1,0 +1,1 @@
+$duration-3ms: 0.3s;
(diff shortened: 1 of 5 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$duration-3ms: 0.3s;
```
SanyaBratashchuk, line 1:
```suggestion
$transition-duration: 0.3s;

```

SanyaBratashchuk: 0.3s is not a 3ms it is 300ms

SanyaBratashchuk (approved):

PR #1232, opened 2022-11-12
catalog_hover

Lav1004:
- [DEMO LINK](https://Lav1004.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -11,7 +11,8 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
+    perspective: 400px;
+    transition-duration: 300ms;
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/header.scss
```diff
@@ -18,25 +18,32 @@
   &__list {
     display: flex;
     margin: 0;
-    padding: 0;
+    padding: 0px;
```
TarasHoliuk, line 21: redundant change, leave it as it was before

src/styles/blocks/catalog.scss
```diff
@@ -11,7 +11,8 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
+    perspective: 400px;
+    transition-duration: 300ms;
```
TarasHoliuk, line 15: set `transition-property` as well, it is better than default `all`
[transition-property](https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property)

TarasHoliuk (changes requested):

tetiana-ivanchenko (changes requested):
fix comment about transition

PR #1236, opened 2022-11-12
Hoverrrr

IoghiStudio:
- [DEMO LINK](https://IoghiStudio.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 300ms, transform 300ms;
+
```
Alireza-Mohammadhossein, line 21: you can use "transition: all 0.3s" for all properties

Alireza-Mohammadhossein (commented):
well done bro

src/styles/blocks/catalog.scss
IoghiStudio: Thanks! I did not know that

naziamoff: actually, that is considered to be a bad practice to use `transition: all`
We usually recommend creating a separate variable for the `transition-duration` and using a specific rule.
`
transition-duration: $transition-duration;
`

PR #1238, opened 2022-11-12
initial commit

olena-yanovska:
- [DEMO LINK](https://olena-yanovska.github.io/layout_catalog_hovers/)


RicksonBro (approved):
GJ! The only thing i'd add it's to put some color values and size values to variables.

olena-yanovska:
> GJ! The only thing i'd add it's to put some color values and size values to variables.

Thanks! I don't think that color and size values should be added as variables, because they are used just once in the project

Ihor-Karpyn (approved):

PR #1249, opened 2022-11-13
first version

Vladimir-Zadorozhnyi:
https://Vladimir-Zadorozhnyi.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/utils/variables.scss
+++ b/src/styles/utils/variables.scss
@@ -1,0 +1,2 @@
+$time: 0.3s;
+$smooth: ease;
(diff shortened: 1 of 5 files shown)


src/styles/utils/variables.scss
```diff
@@ -0,0 +1,2 @@
+$time: 0.3s;
```
Anastasiia-Svintsova, line 1: it's better to call it `$duration`

Anastasiia-Svintsova (approved):

PR #1252, opened 2022-11-13
Develop

dolnaro:
- [DEMO LINK](https://dolnaro.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform 0.3s, color 0.3s;
```
Edriss572, line 21: you use transition duration few time if we want you you would use variable

naziamoff (approved):
Well done!
But pls consider the comment

PR #1253, opened 2022-11-13
Develop

Vlad-Dihtiar:
- [DEMO LINK](https://Vlad-Dihtiar.github.io/layout_catalog_hovers/)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $duration;
```
Bohdan-Maksymiuk, line 41: You should specify transition properties directly

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,4 @@
+$main-color-text: #060b35;
+$main-color-accent: #00acdc;
+$title-hover: #34568b;
```
Bohdan-Maksymiuk, line 3: it should be clear from name that it is color

Bohdan-Maksymiuk (changes requested):

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,4 @@
+$main-color-text: #060b35;
+$accent-color: #00acdc;
+$title-card-hover: #34568b;
```
polosanya, line 3:
```suggestion
$title-hover-color: #34568b;
```

polosanya (approved):
Good job!

PR #1255, opened 2022-11-14
create hovers

Andrey-Ivanov999:
- [DEMO LINK](https://Andrey-Ivanov999.github.io/layout_catalog_hovers/)
-


src/styles/blocks/catalog.scss
```diff
@@ -1,3 +1,5 @@
+$transition-duration:300ms;
```
YakovlewDan, line 1: I think it must be in separate file for variables.

src/styles/blocks/catalog.scss
```diff
@@ -1,3 +1,5 @@
...
+
 .catalog {
```
YakovlewDan, line 3: Same case with separate files, there must be a folder for blocks. 
By the way, you forgot about 'header' block, check p4 of the main task (Smoothly (during 300ms) change navigation link text color to #00acdc)

YakovlewDan (changes requested):
BEM structure not adjusted(checklist p2)

src/styles/blocks/catalog.scss
Andrey-Ivanov999: Done

valerii-derkach (approved):

PR #1257, opened 2022-11-14
commit message

miftaha:
[DEMO LINK](https://miftaha.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,14 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms transform, 300ms color;
```
nurdanokcu, line 20:
```suggestion
    transition: transform 300ms , color 300ms ;
```

src/styles/blocks/header.scss
```diff
@@ -26,10 +26,15 @@
     color: #060b35;
     list-style: none;
     line-height: 60px;
+    transition: 300ms color;
```
nurdanokcu, line 29:
```suggestion
    transition: color 300ms ;
```

src/styles/blocks/catalog.scss
nurdanokcu: It is better to put transition property name first then other attributes.
You can read more about it [here](https://developer.mozilla.org/en-US/docs/Web/CSS/transition)

nurdanokcu (changes requested):

nurdanokcu (approved):
Everything is as expected. ✅

naziamoff (approved):
🔥

PR #1258, opened 2022-11-14
Develop

Kavster92:
- [DEMO LINK](https://Kavster92.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,12 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+
+    transition: 300ms;
+
+    &:hover {
+      color: $color-blue;
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: 300ms;
```
miftaha, line 42:
```suggestion
    transition: 300ms;
```
 transition: color 300ms || transition: color $duration

miftaha (changes requested):

src/styles/blocks/header.scss
naziamoff: I agreee with @miftaha, it is considered to be a better practice to write the transition property explicitly

naziamoff (approved):
GJ! But take a look at my comment

PR #1264, opened 2022-11-16
Develop

nazar-medushevskyi:
- [DEMO LINK](https://nazar-medushevskyi.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_ulits.scss
+++ b/src/styles/_ulits.scss
@@ -1,2 +1,3 @@
 $color--hover--link: #00acdc;
 $color--hover--text: #34568b;
+$transitionSpeed : 300ms;


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
+  }
```
Bohdan-Maksymiuk, line 42: You should specify transition properties directly

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 0.3s, color 0.3s;
```
Bohdan-Maksymiuk, line 20: same comment here

Bohdan-Maksymiuk (changes requested):
Good job, lets fix comments

src/styles/blocks/header.scss
nazar-medushevskyi: you mean not using duration, and in the second instead of transition: transform 0.3 s, color 0.3 s; write a simple transition: 0.3s; ?

nazar-medushevskyi:
I added a value that will be responsible for the transition time, now everything is simpler and when I change the value in one place, everything is rebuilt for it, now I understand the error. Thanks

nazar-medushevskyi:
Снимок экрана 2022-11-17 в 14 34 15
почему то не показывается что я поменял , вот скрин кода после отправки

src/styles/_ulits.scss
```diff
@@ -0,0 +1,2 @@
+$color--hover--link: #00acdc;
+$color--hover--text: #34568b;
```
polosanya, line 2: Add newline in the end
https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $transitionSpeed, color $transitionSpeed;
```
polosanya, line 20: I can't find a declaration of these vars. Probably you forgot to add file with them to your commit or to deploy the changes

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $transitionSpeed;
```
polosanya, line 41: You should specify for which properties you want to apply the transition. 
Because currently it applies to all the properties and it's not good for optimization.
I guess Bohdan's comment was about that.

polosanya (changes requested):
Almost perfect! Please check the comments

SanyaBratashchuk (approved):

PR #1265, opened 2022-11-17
Develop

nadia-hope:
- [DEMO LINK](https://nadia-hope.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,3 +1,5 @@
+$hover-text: #34568b;
+
 .catalog {
   display: flex;
   flex-wrap: wrap;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +20,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform 300ms;
```
qoa11a, line 23: Let's add transition for color as well.

qoa11a (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -1,3 +1,5 @@
+$hover-text: #34568b;
```
polosanya, line 1:
```suggestion
$hover-text-color: #34568b;
```
It is better to name it the way it's understandable that it is a color

src/styles/blocks/header.scss
```diff
@@ -1,3 +1,5 @@
+$hover-text: #34568b;
```
polosanya, line 1: Same. Btw, there should be another color for links
image

polosanya (changes requested):
Almost perfect! Please check the comments

loralevitska (approved):
good job!

PR #1275, opened 2022-11-22
I can't import files from utils folder

RuslanLiapin:
I @import "utils/mixins.scss"
@import "utils/variables.scss"
And it doesn't work
Error can't find variables


src/styles/blocks/catalog.scss
```diff
@@ -1,8 +1,15 @@
+@mixin card-size {
+  width: 200px;
+  height: 300px;
+}
+$catalog-max-width: 944px;
+$margin: 48px;
```
liliia-livandovskaya, line 6: it should be removed from this file

src/styles/blocks/header.scss
```diff
@@ -1,3 +1,5 @@
+$hover-color-nav: #00acdc;
```
liliia-livandovskaya, line 1: the same here, you already have file with variables

liliia-livandovskaya (changes requested):
great job! Pay attention to the comments

Viktor-Kost (approved):

PR #1279, opened 2022-11-22
Hovers was added

hyrych-max:
[DEMO LINK] - https://hyrych-max.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -11,6 +11,7 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
+    cursor: pointer;
 
     width: 200px;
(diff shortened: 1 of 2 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -21,5 +22,12 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    transition: $transition-time;
```
Anastasiia-Svintsova, line 26: You need to add properties for the transition

Anastasiia-Svintsova (changes requested):
Good! Fix small comment

polosanya (approved):
Well done!

PR #1283, opened 2022-11-25
layout_catalog_hovers

andrii-kovalskyi02:
- [DEMO LINK](https://andrii-kovalskyi02.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -14,12 +14,21 @@
 
     width: 200px;
     height: 300px;
+
+    color: #34568b;
+
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: color $transition-duration, transform $transition-duration;
+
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -14,12 +14,21 @@
 
     width: 200px;
     height: 300px;
+
+    color: #34568b;
```
loralevitska, line 18: it's good practice move all colors to variables

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: color $transition-duration;
+
+    &:hover {
+      color: #00acdc;
```
loralevitska, line 45: same here)

loralevitska (changes requested):
almost done!

SanyaBratashchuk (approved):

PR #1288, opened 2022-11-27
Catalog hovers added

undepend:
- [DEMO LINK](https://undepend.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms, color 300ms;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms, color 300ms;
```
Viktor-Kost, line 20: Add a variable for transition-duration. So you don't need to repeat this value several times

Viktor-Kost (changes requested):
Well done! 
A little improvement needed

Virass (approved):

PR #1292, opened 2022-11-28
Added transition properies for the card and navigation links

kostya-stasenko:
[DEMO LINK](https://<kostya-stasenko>.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all 0.3s ease-in-out;
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
+    }
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all 0.3s ease-in-out;
```
tetiana-ivanchenko, line 20: don't use a transition for all, use only those properties for which you need to add a transition

tetiana-ivanchenko (changes requested):

src/styles/blocks/catalog.scss
kostya-stasenko: Даже если у меня одинаковое значение transition-duration и transition-timing-function для color и transform, мне всё равно нужно прописывать значения для каждого отдельно в transition?

polosanya: It is better to specify only needed properties as `all` can hurt the perfomance

src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 0.3s ease-in-out, transform 0.3s ease-in-out;
```
polosanya, line 20: You can also set it like this if you have many properties with same transition
```suggestion
    transition-property: color, transform;
    transition-duration: 0.3s;
    transition-timing-function: ease-in-out;
```

polosanya (approved):
Good job!

PR #1304, opened 2022-12-03
layout_catalog_hovers created

ZupaFly:
- [DEMO LINK](https://ZupaFly.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $duration;
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
+    }
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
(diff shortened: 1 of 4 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$duration: 300ms;
```
polosanya, line 1: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $duration;
```
polosanya, line 21: Please specify to which properties you want to apply the transition, because currently it applies to all the properties, and that's not good for performance.

polosanya (changes requested):
Almost perfect! Please check the comments to improve your code a bit)

src/styles/blocks/catalog.scss
ZupaFly: I understand my mistakes. Both about new line and transition properties. Changed code according to the comments. Please check

qoa11a (approved):

PR #1311, opened 2022-12-05
Develop

Peti51:
https://Peti51.github.io/layout_catalog_hovers/


src/styles/blocks/catalog.scss
```diff
@@ -11,7 +11,7 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
+    transition-duration: var(--animation-duration);
```
polosanya, line 14: Please specify to which exactly properties you want to apply the transition. Currently, it applies to all the properties and it's not good for optimization(

src/styles/blocks/catalog.scss
```diff
@@ -22,4 +22,8 @@
       margin-right: 48px;
     }
   }
+  &__card:hover {
+    transform: scale(120%);
+    color: #34568b;
```
polosanya, line 27: image
Also, it would be great to create variables for colors

polosanya (changes requested):
Almost perfect! Please check my comments to improve your code a bit

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: var(--animation-duration-header);
```
polosanya, line 41:
```suggestion
    transition: color var(--animation-duration-header);
```
Transition properties are still not specified. Please fix it like that

polosanya (changes requested):
Last small change

src/styles/blocks/catalog.scss
```diff
@@ -11,15 +11,19 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transition var(--anim-time-cat), color var(--anim-time-cat);
```
qoa11a, line 19:
```suggestion
    transition: transform var(--anim-time-cat), color var(--anim-time-cat);
```

src/styles/blocks/variables.scss
```diff
@@ -0,0 +1,6 @@
+:root {
+  --animation-duration-header: 300ms;
+  --anim-time-cat: 300ms;
```
qoa11a, line 3: What is cat? My first thought was that it's about cat as animals. Let's give proper names. For example, --animation-duration-card/catalog/sth else.

qoa11a (changes requested):
Few comments.

jstmpelowycz (approved):
LGTM

PR #1315, opened 2022-12-06
solved the task

MeryamWazeen:
- [DEMO LINK](https://meryamwazeen.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+  }
+
+  &__card:hover {
+    color: #34568b;
+    transform: scale(120%);
   }
 }
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
+  }
+
+  &__link:hover {
+    color: #00acdc;
   }
 }


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
...
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
   }
+
+  &__card:hover {
+    color: #34568b;
+    transform: scale(120%);
+  }
```
jstmpelowycz, line 30:
```suggestion
    transition-duration: 300ms;
    
    &:hover {
      color: #34568b;
      transform: scale(120%);
    }

    &:not(:nth-child(4n)) {
      margin-right: 48px;
    }
  }
```

jstmpelowycz (changes requested):
let's use nesting properly

src/styles/blocks/catalog.scss
MeryamWazeen: done

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
polosanya, line 20:
```suggestion
    transition-duration: 300ms;
    transition-property: color, transform;
```
Would be great to specify to which exactly properties you want to apply the transition. 
If not specified, it applies to all the properties which is not really good for performance.

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
+  }
+
+  &__link:hover {
+    color: #00acdc;
```
polosanya, line 45: Same here. Also, would be great to create variables for main values (like colors and transition time)
image

polosanya (changes requested):
Almost perfect!

src/styles/blocks/catalog.scss
MeryamWazeen: done

src/styles/blocks/header.scss
MeryamWazeen: done

qoa11a (approved):
Good job!

PR #1317, opened 2022-12-06
done

Vadivel-Babu:
[DEMO LINK](https://Vadivel-Babu.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all 0.3s;
```
jstmpelowycz, line 20: do not apply `transition` to all changed properties, select required one only

jstmpelowycz (changes requested):

PR #1332, opened 2022-12-09
finished

barantarasNEW:
https://barantarasNEW.github.io/layout_catalog_hovers/


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,17 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-property: transform, color;
+    transition-duration: 600ms;
```
Viktor-Kost, line 21: It's long duration for hover effect. Use 200-300ms instead

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,17 @@
...
+    transition-timing-function: linear;
```
Viktor-Kost, line 22: `linear` is a default timing-function, no needs to set it

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 300ms linear;
```
Viktor-Kost, line 41: Better to create a variable for duration and apply both to header and catalog hovers

Viktor-Kost (changes requested):

Viktor-Kost (approved):
👍

PR #1340, opened 2022-12-11
v1

GUSILLUS:
- [DEMO LINK](https://GUSILLUS.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition:all 0.3s ease;
```
ivanchuk-yurii, line 21: don't animate all, specify what do you want to

ivanchuk-yurii (changes requested):

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    
+    transition: 0.3s;
```
witflash, line 42: If you want to set duration only, use `transition-duration` rule.

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,17 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: 
+    transform 0.3s ease, 
+    color 0.3s ease;
```
witflash, line 23:
```suggestion
    transition: 
      transform 0.3s ease, 
      color 0.3s ease;
```
It's better to add indentation.

witflash, line 23: Also, you could move `transition-duration` to CSS variables to avoid value duplication.

witflash (changes requested):
Good job!
Pay attention to the comments.

OlEzhA0 (changes requested):
Screenshot 2022-12-16 at 5 38 39 PM
please check why your page isn't deployed

tetiana-ivanchenko (approved):

PR #1350, opened 2022-12-12
first commit catalog hover

Spec0s:
https://spec0s.github.io/layout_catalog_hovers/


src/styles/blocks/catalog.scss
```diff
@@ -7,6 +7,7 @@
   padding: 50px 0;
 
   &__card {
+    transition: transform 200ms, color 300ms;
```
polosanya, line 10: Do you really need to have different transition duration for different properties?
I think it is better to keep it the same for the whole page in order not to confuse the user)

Spec0s:
Sorry !  I don't know why but I can swear it was required 200ms in description at one place, now I checked and I can see its 300ms everywhere. I must have been tired 😁🤣😒

PR #1359, opened 2022-12-14
added hover styling to cards and header links, added variables list

Oleksandr-Maly:
- [DEMO LINK](https://Oleksandr-Maly.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -11,6 +11,7 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
+    transition: $time;
```
anna-inozemets, line 14: It is better to add property for transition, like transition: color $time; (where color - is property which was changed)

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $time;
+
+    &:hover {
+      color: #00acdc
```
anna-inozemets, line 44: U can create a variable for colors

anna-inozemets (approved):
Really good job)

alexandra-protyanova (changes requested):

src/styles/blocks/header.scss
Oleksandr-Maly: but we use it only once

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $time;
```
alexandra-protyanova, line 41:
```suggestion
    transition: color $time;
```

alexandra-protyanova (approved):

PR #1377, opened 2022-12-14
Develop

chiper9:
- [DEMO LINK](https://chiper9.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: color $transition-duration;
+
+    &:hover {
+      color: #00acdc;
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: color $transition-duration;
+
+    &:hover {
+      color: #00acdc;
```
yaroslavna-horban, line 45: you should use variables for colors

yaroslavna-horban (changes requested):
It needs to be slightly improved, but you did great!

src/styles/blocks/header.scss
chiper9: I thought about it but the checklist says "But don't overuse them, don't create variables for the value that's used just once". So I used it once and didn't create for in that case. But I really appreciate your attention to my work. Thank you,  Yaroslava ;)

alexandra-protyanova: You use it twice actually)

alexandra-protyanova (approved):

PR #1379, opened 2022-12-14
Catalog hovers task

vanst1:
- [DEMO LINK](https://vanst1.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,7 +1,6 @@
 .catalog {
   display: flex;
   flex-wrap: wrap;
-
   max-width: 944px;
   margin: 0 auto;
   padding: 50px 0;
@@ -11,12 +10,17 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
+
+    &:hover {
+      transform: scale(120%);
+      color: #34568b;
+    }
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -11,12 +10,17 @@
...
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
+
+    &:hover {
+      transform: scale(120%);
+      color: #34568b;
+    }
```
GarryRocksman, line 23: you need to set not only duration but with styles that will be modified by the transition. Like `transition: color , transform`  [read](https://developer.mozilla.org/ru/docs/Web/CSS/transition)

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
+
+    &:hover {
+      color: #00acdc;
+    }
```
GarryRocksman, line 45: here the same
Screenshot 2022-12-14 at 20 35 00

GarryRocksman (changes requested):
You did great work. Just go through my comments and see if you need some fixes.  Also, I recommend creating variables for setting colors, and maybe time duration

src/styles/blocks/catalog.scss
```diff
@@ -11,12 +10,17 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 300ms, transform 300ms;
```
Denchepornyuk, line 18: create scss varibale for `300ms`

Denchepornyuk (changes requested):

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$transition-duration: 300ms;
```
Denchepornyuk, line 1: add empty line at the end of the file

Denchepornyuk (approved):

PR #1380, opened 2022-12-14
catalog-hover

OlehSharha:
- [DEMO LINK](https://OlehSharha.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,4 @@
+@import "../styles/utilit/variable.scss";
 @import "blocks/page";
 @import "blocks/header";
 @import "blocks/catalog";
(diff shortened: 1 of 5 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,4 @@
+@import "../styles/utilit/variable.scss";
```
maasiimka, line 1: Your main file is on the same level as the `utilit` folder, so you don't need this `../styles/`
Also, `.scss` is redundant there
```suggestion
@import "utilit/variable";
```

maasiimka (changes requested):
You have got some mistakes to fix, but I think that you did a great job! Good luck ;)

Denchepornyuk (approved):

PR #1381, opened 2022-12-14
catalog hovers commit

serbinm:
- [DEMO LINK](https://serbinm.github.io/layout_catalog_hovers/)


src/styles/main.scss
```diff
@@ -1,3 +1,7 @@
-@import "blocks/page";
-@import "blocks/header";
-@import "blocks/catalog";
+// @import "blocks/page";
+// @import "blocks/header";
+// @import "blocks/catalog";
```
GarryRocksman, line 3: remove comments

GarryRocksman (changes requested):
Hi, looks like your code is ok, but the demo link doesn't work. And you have too many files committed. There are several that you don't need
Screenshot 2022-12-14 at 21 21 06

denys-cheporniuk (approved):

PR #1383, opened 2022-12-14
transitions was added

ihor-soloviov:
- [DEMO LINK](https://ihor-soloviov.github.io/layout_catalog_hovers/)


src/styles/blocks/header.scss
```diff
@@ -26,10 +26,15 @@
     color: #060b35;
     list-style: none;
     line-height: 60px;
+    transition: $transition-time;
```
victorionov, line 29: it's better to indicate which property you animated.

victorionov (approved):

src/styles/blocks/header.scss
Denchepornyuk: agree, add style into transition

denys-cheporniuk (changes requested):

src/styles/blocks/header.scss
ihor-soloviov: ops

Anna-Tkachenko (approved):
Good job!

PR #1399, opened 2022-12-20
Initial commit

Intellium7:


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +19,17 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: width, height, color, 300ms linear;
```
ViktorHavryliuk, line 22: Please use variable for transition time

Also, you may use transform instead of width, height
`transition: transform, color, 300ms linear;`

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: color 300ms linear;
```
ViktorHavryliuk, line 42: Please use same variable here as well

ViktorHavryliuk (changes requested):

src/styles/blocks/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$time: 300ms;
```
tetiana-ivanchenko, line 1:
```suggestion
$transition-duration: 300ms;
```

tetiana-ivanchenko (approved):

agniya-i (approved):

PR #1403, opened 2022-12-21
catalog was changed

KirillMaslov:
[DEMO LINK](https://KirillMaslov.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +18,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
ViktorHavryliuk, line 21: Please use variable for transition time

It's better to aslo specify transition property (color, transform, etc.) directly, please add an transition-property rule, or just use shorthand like transition: transform $transition-time

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
ViktorHavryliuk, line 41: Same here

ViktorHavryliuk (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +18,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $transition-time;
```
ViktorHavryliuk, line 21: Specify transition-property directly

ViktorHavryliuk (changes requested):

tetiana-ivanchenko (approved):

PR #1407, opened 2022-12-23
Develop

serhiilysenko:
DEMO LINK https://serhiilysenko.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 0.3s;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      color: #34568b;
+      transform: scale(120%);
     }
   }
 }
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 0.3s;
```
ViktorHavryliuk, line 20: Please use variable for transition time as you use it in few places
Also please specify which proprty you are transtioning, using transition-property rule or just transition rule with several params, like transition: transform, color $transition time;

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 0.3s;
```
ViktorHavryliuk, line 41: Same here

ViktorHavryliuk (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color transform $hover-transition;
```
TarasHoliuk, line 20:
```suggestion
    transition: color, transform $hover-transition;
```
now it's broken

TarasHoliuk (changes requested):

ViktorHavryliuk (approved):

PR #1416, opened 2022-12-26
catalog-hover

p0mk:
ttps://p0mk.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -5,6 +5,7 @@
   max-width: 944px;
   margin: 0 auto;
   padding: 50px 0;
+  perspective: 1em;
 
   &__card {
     display: flex;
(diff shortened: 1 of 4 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +19,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $speed-transition;
```
DarkMistyRoom, line 22: list all transition properties, it is better for performance

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $speed-transition;
```
DarkMistyRoom, line 41: list all transition properties, it is better for performance

DarkMistyRoom (changes requested):

nazarmatsevych (approved):

PR #1422, opened 2022-12-28
added transitions

domkab:
- [DEMO LINK](https://domkab.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -12,6 +12,9 @@
     align-items: center;
     box-sizing: border-box;
 
+    transition-duration: $transition-duration;
+    transition-property: transform, color;
+
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
@@ -20,6 +23,12 @@
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      transform: scale(120%);
+
+      color: #34568b;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -21,5 +24,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(120%);
+
+      color: #34568b;
```
polosanya, line 31: Would be great to create variables for colors too)

src/styles/blocks/header.scss
```diff
@@ -33,10 +33,15 @@
   }
 
   &__link {
+    transition: $transition-duration;
```
polosanya, line 36: Please specify to which properties you want to apply the transition, as you did in catalog.scss
For now it applies to all the properties and it's not good for performance

polosanya (changes requested):
Almost perfect!

src/styles/blocks/header.scss
```diff
@@ -33,10 +33,17 @@
...
```
polosanya, line 36:
```suggestion
    transition-duration: $transition-duration;
```
You don't need to use shorthand property now as it controls only duration.
Or you can set both of them like this:
`transition: color $transition-duration`

polosanya (approved):
Well done! Please pay attention to my comment

PR #1424, opened 2022-12-28
added hover effects

venher-v:
- [DEMO LINK](https://venher-v.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/cards.scss
+++ b/src/styles/blocks/cards.scss
@@ -1,0 +1,79 @@
+.card {
+  padding: 32px 16px 16px;
+  background-color: $white-color;
+  border: $card-border solid $white-smoke-color;
+  border-radius: 5px;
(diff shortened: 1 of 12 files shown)


qoa11a (changes requested):
When I hover over card text becomes blue and then again returns to black even though I'm still hovering over the card.

src/styles/blocks/cards.scss
```diff
@@ -0,0 +1,77 @@
+.card {
+  padding: 32px 16px 16px;
+  background-color: $white-color;
+  border: $card-border solid $white-smoke-color;
+  border-radius: 5px;
+  transition: color 0.3s, transform 0.3s;
```
SanyaBratashchuk, line 6: move 0.3s to the variable

SanyaBratashchuk (approved):

PR #1433, opened 2023-01-04
catalogHovers1

edkido:
- [DEMO LINK](https://edkido.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms linear;
```
Viktor-Kost, line 20: add exact transition-properties, since it uses `all` by default

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 300ms linear;
```
Viktor-Kost, line 41: add transition-property

src/styles/blocks/catalog.scss
Viktor-Kost, line 20: create a variable for transition-duration

Viktor-Kost (changes requested):

marina-tilniak (approved):

PR #1446, opened 2023-01-10
Task is implemented

kateryna-tilishevska:
- [DEMO LINK](https://kateryna-tilishevska.github.io/layout_catalog_hovers/)


OlEzhA0 (changes requested):
Hey! other Cards shouldn't be affected on hover
Screenshot 2023-01-11 at 10 54 51 AM
`Smoothly (during 300ms) increase the card by 20 percent on hover (neighboring cards should not be affected)`

mykhalenych (approved):

PR #1447, opened 2023-01-10
task is ready

EkaterinaBessonnaya:


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
...
+    transition: all $time;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      color: #34568b;
+      transform: scale(1.2);
+      cursor: pointer;
```
Anastasiia-Svintsova, line 29: Move `cursor: pointer;` from hover

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all $time;
```
Anastasiia-Svintsova, line 20: Don't use all here, specify the properties

Anastasiia-Svintsova (changes requested):

src/styles/blocks/catalog.scss
EkaterinaBessonnaya: where move cursor? don't und...  are just remove it?

qoa11a (changes requested):
You should have moved `cursor: pointer` from :hover to a class.

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color $time, transform $time;
```
polosanya, line 20:
```suggestion
    transition-property: color, transform;
    transition-duration: $time;
```
Just for your information, it can also be written as follow.
In your case it's okay, but if you have a lot of properties, then it's better to use the following variant

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
...
+    cursor: pointer;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      color: #34568b;
```
polosanya, line 28: Would be great to create variables for all the colors too in order to be able to change them in one place

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comments)

PR #1456, opened 2023-01-13
catalog hovers

YelyzavetaMolochynska:
- [DEMO LINK](https://YelyzavetaMolochynska.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -22,4 +22,10 @@
       margin-right: 48px;
     }
   }
+
+  &__card:hover {
+    transform: scale(120%);
+    color: #34568b;
+    animation-duration: 300ms;
+  }
 }
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -4,13 +4,13 @@
   justify-content: space-between;
   box-sizing: border-box;
 
-  height: 60px;
+  height: $header-height;
   padding: 0 50px;
 
   box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
 
   &__logo {
-    height: 40px;
+    height: $logo-size;
   }
 }
 
@@ -38,5 +38,9 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    &:hover {
+      color: $color-accent;
+    }
   }
 }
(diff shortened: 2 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -22,4 +22,10 @@
       margin-right: 48px;
     }
   }
+
+  &__card:hover {
+    transform: scale(120%);
+    color: #34568b;
+    animation-duration: 300ms;
+  }
 }
```
IvanFesenko, line 31: Move this to card element using &:hover

IvanFesenko (changes requested):
Small fix should be done

tetiana-ivanchenko (changes requested):
scaling should be smooth, now it is not

src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transition: 300ms;
+      transform: scale(120%);
+      color: #34568b;
+    }
```
Viktor-Kost, line 29: at the `hover` state you need only set what should be changed
`transition` rule remove to the &__card

src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transition: 300ms;
```
Viktor-Kost, line 26: create a variable for 300ms, because you use it several times

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    &:hover {
+      color: $color-accent;
+      transition: 300ms;
+    }
```
Viktor-Kost, line 45: check comments above and apply similar solution here

Viktor-Kost (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,10 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(120%);
```
Anastasiia-Svintsova, line 26:
```suggestion
      transform: scale(1.2);
```

Anastasiia-Svintsova (approved):

PR #1462, opened 2023-01-15
develop

Roman-Kulchytskyi:
https://roman-kulchytskyi.github.io/layout_catalog_hovers/


src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
+      transition: 300ms;
```
tetiana-ivanchenko, line 28: transition shouldn't be on hover, use it directly to the element, because if you move mouse out, transition don't works.
also use transition-property

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    &:hover {
+      color: #00acdc;
+      transition: 300ms;
```
tetiana-ivanchenko, line 44: the same

tetiana-ivanchenko (changes requested):

Viktor-Kost (approved):

PR #1465, opened 2023-01-16
completed task

NikiforovSG:
[DEMO LINK](https://NikiforovSG.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform 0.3s, color 0.3s;
+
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform 0.3s, color 0.3s;
```
marina-tilniak, line 21: move 0.3s to variable

marina-tilniak (changes requested):
Would  be good to fix it and move cards to the center of page 
image

vlad-demchuk (approved):

PR #1468, opened 2023-01-16
Transform

HannaVasylieva:
- [DEMO LINK](https://HannaVasylieva.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: color 300ms, transform 300ms;
```
Viktor-Kost, line 21: create a variable for 300ms, because you have already used it twice

src/styles/blocks/header.scss
```diff
@@ -19,6 +19,7 @@
     display: flex;
     margin: 0;
     padding: 0;
+    transition: 300ms;
```
Viktor-Kost, line 22: add transition-property. By default it's `"all"`. You should specify what exactly you would like to change

witflash (approved):
Well done 👍

PR #1476, opened 2023-01-18
Hovers

Vasyl-Zhyliakov:
https://Vasyl-Zhyliakov.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,10 +18,11 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform $smooth-on-hover, color $smooth-on-hover;
+
     &:hover {
       transform: scale(1.2, 1.2);
       color: #34568b;
-      transition: transform $smooth-on-hover, color $smooth-on-hover;
     }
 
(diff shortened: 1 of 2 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,12 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    &:hover {
+      transform: scale(1.2, 1.2);
+      color: #34568b;
+      transition: transform $smooth-on-hover, color $smooth-on-hover;
```
qoa11a, line 24: you should add transition on whole class rather than on hover, so transition works both ways

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    &:hover {
+      color: #00acdc;
+      transition: color $smooth-on-hover;
```
qoa11a, line 44: same here

qoa11a (changes requested):

SanyaBratashchuk (approved):

PR #1484, opened 2023-01-22
Catalog__hovers

SashaKozoriz:
- [DEMO LINK](https://SashaKozoriz.github.io/layout_catalog_hovers/)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,104 @@
+.card {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  width: $card-width;
+  padding: 32px 16px 16px;
+  background: $color-white;
+  border: 1px solid #f3f3f3;
+  border-radius: 5px;
+  transition-duration: 300ms;
```
polosanya, line 10: Add also a transition-property or use a shorthand transition property as follow:
`transition: transform 300ms`
For now the transition applies to all the properties which is not good for the performance

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  &__name {
+    margin-top: 40px;
+    margin-bottom: 4px;
+    font-weight: 500;
+    font-size: 12px;
+    line-height: 18px;
+    color: $color-dark;
+    transition-duration: 30ms;
```
polosanya, line 28: Are you sure that it should be so fast? 
Let's create and use a variable for transition duration to make it the same per page

src/styles/blocks/card.scss
```diff
@@ -0,0 +1,104 @@
...
+
+    &:hover {
+      color: #34568b;
```
polosanya, line 31: Also, would be great to add variables for all the colors

polosanya (changes requested):
Almost done!

Anastasiia-Svintsova (approved):

PR #1487, opened 2023-01-23
first commit

artem-shuhailo:
- [DEMO LINK](https://artem-shuhailo.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,7 +17,7 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition: transform $transition-duration;
+    transition: transform $transition-duration, color $transition-duration;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $transition-card-duration;
```
Anastasiia-Svintsova, line 41: You should specify  the transition property
```suggestion
    transition: color $transition-card-duration;
```

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $transition-card-duration;
```
Anastasiia-Svintsova, line 20: Same

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$transition-card-duration: 300ms;
```
Anastasiia-Svintsova, line 1: You can shorten it, but it's up to you
```suggestion
$transition-duration: 300ms;
```

Anastasiia-Svintsova (changes requested):
Good!  Check my comments

src/styles/blocks/catalog.scss
artem-shuhailo: Here I have both transform and color. Should I write transition: transform $transition-duration; ? As when I write both transform and color - it doesn't work

agniya-i (approved):

PR #1490, opened 2023-01-23
Task is done

sergik94:
https://sergik94.github.io/layout_catalog_hovers/


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,9 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 300ms;
+    &:hover {
```
Viktor-Kost, line 42: add a transition-property
```suggestion
    transition: 300ms;
    
    &:hover {
```

src/styles/blocks/page.scss
```diff
@@ -1,3 +1,5 @@
+$trans-dur: 300ms;
```
witflash, line 1: It's better to use full words for variable names: `$transition-duration`. It will be more readable for other teammates in your team.

witflash (approved):
Well done 👍

PR #1491, opened 2023-01-23
added hover transition

Luba-Bykova:
- [DEMO LINK](https://Luba-Bykova.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,10 +1,11 @@
+
 .catalog {
   display: flex;
   flex-wrap: wrap;
-
-  max-width: 944px;
+  max-width: $max-width;
   margin: 0 auto;
   padding: 50px 0;
+  transition: max-width 300ms 5ms;
 
   &__card {
     display: flex;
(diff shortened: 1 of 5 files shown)


DarkMistyRoom (changes requested):
image

qoa11a (changes requested):
see the same errors mentioned before (other cards shouldn't be affected when user hovers any card)

tetiana-ivanchenko (changes requested):
on your Demo link transition doesn't works, mb you didn't deploy?

src/styles/blocks/catalog.scss
```diff
@@ -1,25 +1,33 @@
...
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color $card-time, scale $card-time;
```
tetiana-ivanchenko, line 20: you change transform and color, but add transition for scale, fix it

tetiana-ivanchenko (changes requested):

ViktorHavryliuk (approved):

PR #1500, opened 2023-01-27
solution 

sich-mykhailo:


src/styles/blocks/header.scss
```diff
@@ -25,18 +25,23 @@
...
-    line-height: 60px;
+    line-height: $header-height;
 
     &:not(:last-child) {
       margin-right: 20px;
     }
   }
 
   &__link {
+    transition-duration: 300ms;
```
vkryvytskyy, line 36: define property and timing-function

vkryvytskyy (approved):

PR #1501, opened 2023-01-27
Maked all the changes smooth on hover

OlegShokun:
- [DEMO LINK](https://olegshokun.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,6 +17,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $duration, color $duration;
+
+    &:hover {
+      cursor: pointer;
+      transform: scale(120%);
+      color: $color-titleText;
+    }
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $duration, color $duration;
+
+    &:hover {
+      cursor: pointer;
```
ViktorHavryliuk, line 23: Move cursor: pointer to __card itself

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color $duration;
+
+    &:hover {
+      cursor: pointer;
```
ViktorHavryliuk, line 44: Here too, pointer should be in class, not pseudo-class

ViktorHavryliuk (changes requested):

src/styles/blocks/catalog.scss
OlegShokun: done

src/styles/blocks/header.scss
OlegShokun: done

PR #1508, opened 2023-01-30
Solution

Ihor-Zakharov:
https://ihor-zakharov.github.io/layout_catalog_hovers


Volodymyr-Melnyk-Fron10der (changes requested):
Cool, just set transition property as well
image

ViktorHavryliuk (approved):

PR #1511, opened 2023-01-30
commit message

RostyslavKuts:
- [DEMO LINK](https://RostyslavKuts.github.io/layout_catalog_hovers/)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition-duration: 0.3s;
```
TarasHoliuk, line 42: FYI: it's always better (for performance) to define transition-property (as you did in catalog.scss)

TarasHoliuk (approved):

PR #1513, opened 2023-01-31
Catalog hovers and transitions were added

KuharchukDmytro:
https://kuharchukdmytro.github.io/layout_catalog_hovers/


milenadiamonds1 (approved):
Good job! But better add variable for transition duration  because u use it more than 1 time in many files)

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: color 300ms, transform 300ms;
```
illiarerun, line 21: agree with @milenadiamonds1 , add variable for duration

illiarerun (approved):

PR #1514, opened 2023-01-31
message

Sergej161995:
[DEMO LINK](https://Sergej161995.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 0.3s;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 0.3s;
```
vlad-demchuk, line 41: add transition for transform as well. move duration to a variable

vlad-demchuk (changes requested):

nazarmatsevych (approved):

PR #1527, opened 2023-02-01
Task is done

MarkBabich:
- [DEMO LINK](https://MarkBabich.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,7 +1,6 @@
 .catalog {
   display: flex;
   flex-wrap: wrap;
-
   max-width: 944px;
   margin: 0 auto;
   padding: 50px 0;
@@ -11,12 +10,18 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
-    width: 200px;
-    height: 300px;
+    width: $card-width;
+    height: $card-height;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all $transition-duration ease-in-out;
+
+    &:hover {
+      scale: 1.2;
+      color: $title-hover-color;
+      cursor: pointer;
+    }
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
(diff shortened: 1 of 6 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,5 @@
+$card-height: 300px;
+$card-width: 200px;
+$transition-duration: 300ms;
+$title-hover-color: #34568b;
+$link-hover-color: #00acdc;
```
valeriiayarzhemska, line 5: look at the last item in the checklist - use variables for the main values so that you'll be able to reuse them and give them descriptive names. But don't overuse them, don't create variable for the value that's used just once.
maybe its better not to use these variables (except transition one)

src/styles/main.scss
```diff
@@ -1,3 +1,4 @@
-@import "blocks/page";
-@import "blocks/header";
-@import "blocks/catalog";
+@import "blocks/page.scss";
+@import "utils/_variables.scss";
```
valeriiayarzhemska, line 2: as i remember its better to put files from utils on the top

src/styles/blocks/catalog.scss
```diff
@@ -11,12 +10,18 @@
...
     box-sizing: border-box;
-
-    width: 200px;
-    height: 300px;
+    width: $card-width;
+    height: $card-height;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all $transition-duration ease-in-out;
```
valeriiayarzhemska, line 18: one time my work was rejected cuz of "all" in transition, mentor said that u need to write every property

valeriiayarzhemska (changes requested):
GJ, Mark! Just some little things to fix:)

src/styles/utils/_variables.scss
MarkBabich: Totally agree. Overused. Fixed

src/styles/main.scss
MarkBabich: Right. Fixed

src/styles/blocks/catalog.scss
MarkBabich: Oh, then in the future I will do so. Fixed

valeriiayarzhemska (approved):

Andr1yK (approved):

PR #1530, opened 2023-02-01
Catalog hovers added

nazarovoleksiy:
https://nazarovoleksiy.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all 0.3s ease-in-out;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      transform: scale(1.2);
+      color: $c-blue-card-hover;
     }
   }
 }
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -27,8 +27,15 @@
     list-style: none;
     line-height: 60px;
 
+    transition: color 0.3s, transform 0.3s;
+
     &:not(:last-child) {
       margin-right: 20px;
+    }
+
+    &:hover {
+      color: $c-blue-title-hover;
+      transform: color;
     }
   }
 
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # Frontend practice with catalog page and hovers
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_catalog_hovers/)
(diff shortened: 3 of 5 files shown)


src/styles/blocks/header.scss
```diff
@@ -27,9 +27,16 @@
     list-style: none;
     line-height: 60px;
 
+    transition: color 0.3s, transform 0.3s;
```
YaroslavDzyadiv, line 30: I think you should create an extra variable for transition time cause u use it several times.

YaroslavDzyadiv (commented):
Great work!
As for me i would add a different _reset.scss file to reset margins and add box-sizing rule there as well.

YaroslavDzyadiv (approved):
Great work!
I think you should add a different _reset.scss file to reset margins and add box-sizing rule there as well.
Also i think it would be better if you will add your .nav block to a separate file.
Your .nav__list don't have paddings by default, you can delete this rule.

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all 0.3s ease-in-out;
```
illiarerun, line 21: don't use `all` with transition, you should specify every property

src/styles/blocks/header.scss
illiarerun: Agree, add var for duration

illiarerun (changes requested):

src/styles/blocks/catalog.scss
nazarovoleksiy: Fixed.

src/styles/blocks/header.scss
```diff
@@ -27,9 +27,16 @@
...
 
+    transition: color $t-transition-time ease-in-out;
+
     &:not(:last-child) {
       margin-right: 20px;
     }
+
+    &:hover {
+      color: $c-blue-title-hover;
+      transform: color;
```
Yuliia-Hrinchenko1, line 38: I think this is redundant code  transform: color;

Yuliia-Hrinchenko1 (approved):

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform $t-transition-time, color $t-transition-time ease-in-out;
```
sergii-nosachenko, line 21: Add transition function to variables too, if you use it everywhere.

src/styles/blocks/header.scss
sergii-nosachenko: Fix your mate's comment.

sergii-nosachenko (changes requested):

Andr1yK (approved):

PR #1534, opened 2023-02-02
Catalog hover

denis-bayazitov:
- [DEMO LINK](https://denis-bayazitov.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: transform $smooth, background-color $smooth;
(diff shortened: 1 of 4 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: transform $smooth, background-color $smooth;
```
nazarmatsevych, line 41: But you didnt change transform on hover, so why do you need it here?

nazarmatsevych (changes requested):

nazarmatsevych (approved):

PR #1535, opened 2023-02-02
transitions added

timoxa2t:
- [DEMO LINK](https://timoxa2t.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,6 +17,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-property: transform, color;
+    transition-duration: 300ms;
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-property: transform, color;
+    transition-duration: 300ms;
```
Ambulence, line 21: I like how you solved it, looking better and shorter than solution in one line. Probably could save duration as variable, because you use it twice in code

Ambulence (approved):
Good job and cool solution!

src/styles/blocks/catalog.scss
illiarerun: agree, better create a variable for duration, you want your transitions to be consistent throughout the site

illiarerun (approved):

PR #1537, opened 2023-02-02
Catalog hover

VadKol:
- [DEMO LINK](https://VadKol.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -21,5 +21,12 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    transition: color 300ms linear, transform 300ms linear;
+
+    &:hover {
+      color: #34568b;
+      transform: scale(1.2);
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,12 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    transition: color 300ms linear, transform 300ms linear;
```
arkushenko-anna, line 25: Great job! There is one recommendation for your code) Maybe you can create a variable with transition time because in all of the cases, we use the same time)

arkushenko-anna (changes requested):

arkushenko-anna (approved):

src/styles/blocks/catalog.scss
VadKol: Fixed ✔

sergii-nosachenko (approved):
Good work 👍

PR #1538, opened 2023-02-02
animotions on hover were added for nav__link and catalog__card elements

andriibezkorovainyi:
- [DEMO LINK](https://andriibezkorovainyi.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,18 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    cursor: pointer;
+    transition:
+      $transform-transition,
+      $color-transition;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
     }
   }
 }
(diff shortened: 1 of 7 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,3 @@
+// Colors
```
stanislavcodes, line 1:
```suggestion
```

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,3 @@
...
+$color-transition: color 300ms;
+$transform-transition: transform 300ms;
```
stanislavcodes, line 3: I think you can replace those variables with just one:
`$transition-duration: 300ms;`

And use it like this:
`transition: color $transition-duration, transform $transition-duration;`

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,18 @@
     border-radius: 5px;
     margin-bottom: 48px;
```
stanislavcodes, line 19: You can create a variable for gaps between cards because it's always 48px

stanislavcodes (approved):
Looks good to me!
But variables can be improved

src/styles/utils/_variables.scss
Andr1yK: agree

Andr1yK (approved):

PR #1550, opened 2023-02-03
added layout catalog hovers

ctk12:
[DEMO LINK](https://ctk12.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $transition-duration;
+
+    &:hover {
+      color: $color-accent-blue;
+    }
   }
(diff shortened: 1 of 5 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $transition-duration;
```
polosanya, line 41: Why didn't you specify the property here too as you did above?
Now the transition applies to all the properties, which is not good for performance 

polosanya (changes requested):
Almost done!

src/styles/blocks/header.scss
ctk12: ok sir

qoa11a (approved):

PR #1553, opened 2023-02-03
Animations were added

sasha-dranuta:
- [DEMO LINK](https://sasha-dranuta.github.io/layout_catalog_hovers/)


kuzmi491 (changes requested):
[SASS] - use variables for the main values so that you'll be able to reuse them and give them descriptive names. But don't overuse them, don't create variable for the value that's used just once.

Sorry=)

illiarerun (approved):

PR #1563, opened 2023-02-03
hovers

Natalia1108u:
Replace `<your_account>` with your Github username and copy the links to Pull Request description:
- [DEMO LINK](https://Natalia1108u.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
sergii-nosachenko, line 20: Add transition props, don't rely on default `all` value

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
...
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      cursor: pointer;
```
sergii-nosachenko, line 27: Keep cursor prop with element itself, not its pseudo-class

sergii-nosachenko (changes requested):
Add animation to nav links too (from task):
_Smoothly (during 300ms) change navigation link text color to #00acdc_

src/styles/blocks/catalog.scss
```diff
@@ -5,6 +5,7 @@
   max-width: 944px;
   margin: 0 auto;
   padding: 50px 0;
+  cursor: pointer;
```
qoa11a, line 8: It's incorrect to add `cursor: pointer;` on the entire catalogue. Just remove it from the `:hover` and add it to the `&__card`.

qoa11a (changes requested):
A small fix.

PR #1568, opened 2023-02-03
catalog hover was created

vlad-fil:
- [DEMO LINK](https://vlad-fil.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: $during-hover;
```
IvanFesenko, line 20:
```suggestion
    transition: transform $during-hover color $during-hover;
```

IvanFesenko (approved):
Good job. Keep in mond, that transition by default used for `all`

PR #1576, opened 2023-02-06
adds animation card and link nav

Fedos17:
- [DEMO LINK](https://Fedos17.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,6 +18,8 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform $time-transition, color $time-transition;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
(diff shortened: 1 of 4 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(120%);
+      color: #34568b;
+      transition: 300ms;
```
Viktor-Kost, line 28: 1) move transition to static state instead of hover one
2) add transition-property (-ies) to show what styles you exactly change
3) create a variable for transition-duration in case you've already used this value twice

Viktor-Kost (changes requested):

qoa11a (approved):

PR #1578, opened 2023-02-06
catalog_hovers

SerhiiShapovalenko:
- [DEMO LINK](https://SergeyShapovalenko.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -39,7 +39,7 @@
     font-size: 12px;
     font-weight: 500;
 
-    transition-duration: $transmission-duration;
+    transition: $transition;
 
     &:hover {
       color: #00acdc;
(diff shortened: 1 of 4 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition-duration: $transmission-duration;
```
loralevitska, line 42: Add property to  transition image
Fix everywhere

loralevitska (changes requested):
Almost perfect! Let's improve

src/styles/variables/variables.scss
```diff
@@ -0,0 +1 @@
+$transition: transform 300ms, color 300ms;
```
slavanosatskyi, line 1: minor in real life such variables are rear cases. in your case it was better to move 300ms to variable only. As nav links do not change their transform property on hover event.

slavanosatskyi (approved):
Nice work!

PR #1580, opened 2023-02-06
hover effects were added

Gosyav:
https://vo7kov.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,7 +18,7 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition: transform, 0.3s, color 0.3s;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform, 0.3s, color 0.3s;
```
vlad-demchuk, line 21: move duration to variable as well

vlad-demchuk (changes requested):

Vo7kov:
https://vo7kov.github.io/layout_catalog_hovers/

qoa11a (approved):

PR #1582, opened 2023-02-08
Animation added

khomenkom:
- [DEMO LINK](https://khomenkom.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms;
```
tetiana-ivanchenko, line 20: add transitions property's, also move transition time in the variable

tetiana-ivanchenko (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all $transition-time;
```
tetiana-ivanchenko, line 20: don't add all  property, use the properties you need

tetiana-ivanchenko (changes requested):

khomenkom:
Thanks for your feedback

vlad-demchuk (approved):

PR #1583, opened 2023-02-08
add transitions to cards and nav links

HYPER2307:
- [DEMO LINK](https://hyper2307.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,16 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $transition-duration;
+    transition-timing-function: smooth;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,16 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $transition-duration;
+    transition-timing-function: smooth;
```
VitaliyBondarenko1982, line 22:
```suggestion
    transition: transform $transition-duration smooth;
```

VitaliyBondarenko1982: Better to specify transition propery. It impact on performance. By default it is all.
Please, check in second place it

VitaliyBondarenko1982 (changes requested):

OlEzhA0 (approved):

PR #1602, opened 2023-02-17
solution

andrii-rykal:
[DEMO LINK](https://andrii-rykal.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,7 +1,6 @@
 .catalog {
   display: flex;
   flex-wrap: wrap;
-
   max-width: 944px;
   margin: 0 auto;
   padding: 50px 0;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -11,15 +10,20 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $speed-of-change;
```
nazarmatsevych, line 18: Also add [transition-property](https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property) rule and indicate exactly which properties will change to make better performanse

src/styles/blocks/header.scss
```diff
@@ -38,5 +36,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $speed-of-change;
```
nazarmatsevych, line 39: same here

nazarmatsevych (changes requested):

PR #1605, opened 2023-02-18
make smooth on hover

vitalii-fedusov:
- [DEMO LINK](https://vitalii-fedusov.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +21,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    position: relative;
+    transition: all $duration ease;
```
SanyaBratashchuk, line 25: specify each prop you apply transition for

src/styles/blocks/header.scss
```diff
@@ -38,5 +41,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: all $duration ease;
```
SanyaBratashchuk, line 44: same

SanyaBratashchuk (changes requested):

src/styles/blocks/catalog.scss
vitalii-fedusov: i do not catch it. what i need to change?

mykhalenych (approved):

PR #1612, opened 2023-02-20
Added hover styles and applied scss styles

nugardesq:
https://nugardesq.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/utils/_variables.scss
+++ b/src/styles/utils/_variables.scss
@@ -1,0 +1,10 @@
+$catalog-max-width: 944px;
+$card-hover-color: #34568b;
+$card-hover-scale: 120%;
+$card-height: 300px;
+$card-width: 200px;
+$header-logo-height: 40px;
+$header-height: 60px;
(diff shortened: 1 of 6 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,10 @@
+$catalog-max-width: 944px;
+$card-hover-color: #34568b;
+$card-hover-scale: 120%;
+$card-height: 300px;
+$card-width: 200px;
+$header-logo-height: 40px;
+$header-height: 60px;
+$header-hover-color: #00acdc;
+$nav-item-color: #060b35;
+$transition-duration-time: 300ms;
```
anstsot, line 10: do you really need all these variables? maybe it would be great to add only variables that are used in several places (e.g. `$transition-duration-time`)

anstsot (changes requested):
Looks good! It would be good to check variables and leave only reusable

qoa11a (approved):

PR #1617, opened 2023-02-21
catalog-hovers

Dyakovich-R:
- [DEMO LINK](https://Dyakovich-R.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 300ms;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
     }
   }
 }
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -19,6 +19,8 @@
     display: flex;
     margin: 0;
     padding: 0;
+    transition: 300ms;
+
   }
 
   &__item {
@@ -38,5 +40,9 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    &:hover {
+      color: #00acdc;
+    }
   }
 }


src/styles/blocks/header.scss
```diff
@@ -19,6 +19,8 @@
     display: flex;
     margin: 0;
     padding: 0;
+    transition: 300ms;
```
vlad-demchuk, line 22: move duration to a separate variable to avoid repetitions

vlad-demchuk (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: var(--transition-second);
```
nazarmatsevych, line 20: Also specify properties you change on hover to improve performance

src/styles/blocks/header.scss
```diff
@@ -19,6 +19,8 @@
     display: flex;
     margin: 0;
     padding: 0;
+    transition: var(--transition-second);
```
nazarmatsevych, line 22: same

nazarmatsevych (changes requested):

src/styles/blocks/catalog.scss
Dyakovich-R: I did not understand what exactly I should indicate.

nazarmatsevych: `transition: transform var(--transition-second), color var(--transition-second);`

See more [here](https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property)

If you have any questions - feel free to ask them in fe_chat

src/styles/blocks/header.scss
nazarmatsevych: `transition: color var(--transition-second);`

nazarmatsevych (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform var(--transition-second),color var(--transition-second);
```
nazarmatsevych, line 20:
```suggestion
    transition: transform var(--transition-second), color var(--transition-second);
```

nazarmatsevych (approved):
👍

PR #1622, opened 2023-02-22
added scale and color transitions

ielizarov:
- [DEMO LINK](https://ielizarov.github.io/layout_catalog_hovers/)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: color 300ms;
```
tetiana-ivanchenko, line 42: use your vatiable for the duration

tetiana-ivanchenko (changes requested):

src/styles/blocks/header.scss
ielizarov: Thank you. Already done, please check.

vlad-demchuk (approved):

PR #1634, opened 2023-02-28
add hover effect for cards and nav-links

Valerii05:


src/styles/blocks/catalog.scss
```diff
@@ -11,7 +11,7 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
+    transition: all ease 0.3s;
```
nazarmatsevych, line 14: better to specify properties you are changing on hover insdead of using "all"

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: all ease 0.3s;
```
nazarmatsevych, line 41: same

nazarmatsevych (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -11,7 +11,7 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
+    transition: transform ease 0.3s;
```
nazarmatsevych, line 14: what about color? you also change it on hover

And move transition duration to variable and reuse it everywhere

nazarmatsevych (changes requested):
*

vlad-demchuk (approved):

PR #1638, opened 2023-02-28
added smooth transitions on hover

Esteban4769:
- [DEMO LINK](https://esteban4769.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -7,6 +7,8 @@
   padding: 50px 0;
 
   &__card {
+    @extend %smooth-transition;
+
     display: flex;
     justify-content: center;
     align-items: center;
@@ -20,6 +22,11 @@
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      color: #34568b;
+      transform: scale(1.2);
     }
   }
 }
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -33,10 +33,16 @@
   }
 
   &__link {
+    @extend %smooth-transition;
+
     color: inherit;
     text-decoration: none;
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    &:hover {
+      color: #00acdc;
+    }
   }
 }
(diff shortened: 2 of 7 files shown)


src/styles/blocks/header.scss
```diff
@@ -33,10 +33,16 @@
...
+
     color: inherit;
     text-decoration: none;
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    &:hover {
+      color: #00acdc;
+    }
```
Zad0rnik, line 46: you need to apply this to the item, not to the link, because the hover should not only work when you point at the text itself, but also below and above it

Zad0rnik (approved):
Good job, pay attention to the comment

src/styles/blocks/header.scss
Esteban4769: Its logically wrong to change link style when the user hover on the item, but i get idea about that hover must work on all height of panel, so thanks for suggestion.

src/styles/utils/_extends.scss
```diff
@@ -0,0 +1,3 @@
+%smooth-transition {
+  transition: $transitionDurating linear;
```
VinnieJ-2k20, line 2: Nope, don't fo it this way. You need to specify certain rules that you transition in each case.

VinnieJ-2k20 (changes requested):
Some code changes needed.

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,17 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition:
+      transform 0.3 linear,
```
VinnieJ-2k20, line 21: Move transition duration to variable, it's used several times.

VinnieJ-2k20 (changes requested):
Smooth transition on the card is not working.

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
...
     margin-bottom: 48px;
+    transition: transform $transformTransitionDuration linear,
+      color $colorTransitionDuration linear;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      color: #34568b;
```
SerhiiKirik, line 28: don't forget to create variables for all colors
```suggestion
      color: $c-color;
```

SerhiiKirik (approved):
Great 🔥

PR #1646, opened 2023-03-01
initial commit

glots-a:
https://glots-a.github.io/layout_catalog_hovers/


AnyaLyaLya (approved):
Good job! But I think we didn`t have to add content into cards

VinnieJ-2k20 (approved):
You didn't need to add content to cards, just add transition to what was there.
But, since transition was applied correctly, I'll approve.
Be more attentive next time.

PR #1658, opened 2023-03-01
added some animation to the header and catalog

Sbavdik:


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: 0.3s linear;
```
TanyaS133, line 20: transition: 0.3s - it must be a variable. 
 Because 0.3s is repeated more than 1 time.

src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: $smoothTime linear;
```
VinnieJ-2k20, line 20: Add properties that you transition.

src/styles/blocks/header.scss
```diff
@@ -26,6 +26,11 @@
     color: #060b35;
     list-style: none;
     line-height: 60px;
+    transition: $smoothTime linear;
```
VinnieJ-2k20, line 29: Same here.

VinnieJ-2k20 (changes requested):
Fix transition rule, and that's it!

VinnieJ-2k20 (approved):

PR #1666, opened 2023-03-01
implement hovers

kyserf:
- [DEMO LINK](https://iHateAntichrist.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_variables.scss
+++ b/src/styles/_variables.scss
@@ -1,0 +1,1 @@
+$transitiontime: 300ms;
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,7 +18,7 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition: 300ms;
+    transition: transform $transitiontime, color $transitiontime;
 
     &:not(:nth-child(4n)) {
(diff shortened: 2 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: 300ms;
```
Anna-Kuzmenko, line 21: transition should be in separate file _variables.scss

Anna-Kuzmenko (approved):

src/styles/blocks/catalog.scss
VinnieJ-2k20, line 21: 1. Move transition time to variables, you use it several times.
2. Specify which exact properties are you transitioning.

VinnieJ-2k20 (changes requested):

src/styles/_variables.scss
```diff
@@ -0,0 +1 @@
+$transitiontime: 300ms;
```
Olexandr6, line 1:
```suggestion
$transition-time: 300ms;
```

Olexandr6 (approved):

PR #1675, opened 2023-03-01
added simple animation

Monzolevska:
[DEMO LINK](https://Monzolevska.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 0.3, transform 0.3s;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+    
+    &:hover {
+      color: #34568b;
+      transform: scale(120%);
     }
   }
 }
--- a/src/styles/catalog_container.scss
+++ b/src/styles/catalog_container.scss
@@ -1,0 +1,10 @@
+.container {
+  display: flex;
+  box-sizing: border-box;
+  flex-wrap: wrap;
+  margin: 0 auto;
+  align-items: center;
+  justify-content: center;
+  padding-top: 50px;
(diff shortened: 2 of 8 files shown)


src/styles/product_card.scss
```diff
@@ -0,0 +1,103 @@
+.product {
```
AP-Solution, line 1: You need to add transition to this selector for smoothness effect

src/styles/catalog_container.scss
```diff
@@ -0,0 +1,10 @@
+.container {
+  display: flex;
+  box-sizing: border-box;
```
AP-Solution, line 3: You can replace this property to * selector

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 0.3, transform 0.3s;
```
AP-Solution, line 20: there is an error in this line - add "s" letter after color 0.3

AP-Solution (approved):
Almost perfect! Pay attention to the correctly properties for selectors and reset.css file

SerhiiKirik (changes requested):
Well done 🚀 
Fix bug with hover

src/styles/moyo_header.scss
```diff
@@ -0,0 +1,61 @@
...
+
+    &:not(:last-child) {
+      margin-right: 20px;
+    }
+  }
+
+  &__link {
+    display: inline-block;
+    text-decoration: none;
+    color: var(--main-dark-color);
```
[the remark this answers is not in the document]
Monzolevska: You mean committed all of them? 
I was confused at the beginning so changed to many stuff............ cringe....
and only  after noticed that i pushed all of those......
BTW how do you I put emojis in here
ineed the clown one^ __ ^

liliia-livandovskaya (approved):

PR #1678, opened 2023-03-01
added hover effects according to the task

Olekk17:
- [DEMO LINK](https://olekk17.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,3 +1,5 @@
+$card-transition-duration: 0.3s;
+
 .catalog {
   display: flex;
   flex-wrap: wrap;
@@ -17,6 +19,14 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+
+    transition: transform $card-transition-duration,
+      color $card-transition-duration;
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
+    }
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/header.scss
```diff
@@ -1,3 +1,5 @@
+$header-transition-duration: 0.3s;
```
Olexandr6, line 1: The transition duration should be the same throughout the page. So create a file for variables and put the `transition-duration` there

src/styles/blocks/catalog.scss
```diff
@@ -18,6 +20,14 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform $card-transition-duration,
+      color $card-transition-duration;
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
```
Olexandr6, line 28: If color values are used in several places, makes sense to create variables for them, as well. Check it

Olexandr6 (changes requested):
Almost done👍 Now, let's work on improvements:

Olexandr6 (changes requested):
I don't see any changes from last review, make sure you made git commit + git push

src/styles/blocks/catalog.scss
Olexandr6: Not fixed

Olexandr6 (approved):
Good job 👍
Make sure to create variables for color, as well.

src/styles/blocks/catalog.scss
Olekk17: > I don't see any changes from last review, make sure you made git commit + git push + deploy

We just had to wait a little bit

Olekk17:
> 



> 



> Good job 👍 Make sure to create variables for color, as well.

We use three different colors only for one time each

PR #1682, opened 2023-03-01
FirstVersion

AlexRog01:
- [DEMO LINK](https://AlexRog01.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all $transitionTime;
+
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all $transitionTime;
```
SerhiiKirik, line 21: don't use it like that, because it's the same as transition-duration:  $transitionTime
It's too hard for the browser
Check this
image

```suggestion
    transition: transform $transitionTime, ...;
```

SerhiiKirik (approved):
Well done 🚀

PR #1685, opened 2023-03-02
Hovers

edgarlokys:
https://edgarlokys.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -14,6 +14,7 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
+    cursor: pointer;
     @include card-size;
     border: 1px solid #ccc;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -11,12 +14,17 @@
...
-
-    width: 200px;
-    height: 300px;
+    cursor: pointer;
+    @include card-size;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all $time-hover;
+
```
IvanFesenko, line 23: It's bad pactice to set transition to all. Just specify whta you need
```suggestion
    transition: transform $time-hover, color $time-hover;
```

IvanFesenko (approved):

PR #1689, opened 2023-03-02
add animations for card and nav

StasLukianchuk:
- [DEMO LINK](https://StasLukianchuk.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: 300ms;
+
     &:not(:nth-child(4n)) {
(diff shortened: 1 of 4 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: 300ms;
```
SerhiiKirik, line 21: don't use it like that, because it's the same as transition: all 0.3s
It's too hard for the browser
image
& will be better if you create vars for transition duration & colors
```suggestion
    transition: transform 300ms, color 300ms;
```

SerhiiKirik (approved):
Well done 🚀

PR #1691, opened 2023-03-02
Task completed

KateLis84:
DEMO LINK: https://KateLis84.github.io/layout_catalog_hovers/


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: color 300ms, transform 300ms;
```
vlad-demchuk, line 21: move repeatable values to variables

vlad-demchuk (changes requested):

PR #1703, opened 2023-03-05
Develop

valdesin:


anstsot (changes requested):
Good job, but let's fix some points/
1. use SCSS variables for values, used in several places throughout the code (you can consider using transition time)
2. you've pushed compiled files `style.css` and style`.css.map`. please remove them

PR #1705, opened 2023-03-06
first try

Nikolaymelni:
- [DEMO LINK](https://Nikolaymelni.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $animationDuration;
+
+    &:hover {
+      color: #34568b;
+      transform: scale(1.2);
+    }
+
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $animationDuration;
```
polosanya, line 21: Add `transition-property` properties too
Without it, transition applies to all the properties which is not good for performance

polosanya (changes requested):
Almost perfect!

src/styles/blocks/catalog.scss
Nikolaymelni: Thank you
Done)

agniya-i (approved):

PR #1716, opened 2023-03-09
task complite by Bogdan Pohilko

Bogdan-Pohilko:
- [DEMO LINK](https://Bogdan-Pohilko.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -16,10 +16,17 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    margin-bottom: $card-margin;
+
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -16,10 +16,17 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    margin-bottom: $card-margin;
+
+    transition-duration: 300ms;
```
IvanFesenko, line 21: By default trasition is used for `all`. Need to set what you need
```suggestion
    transition: transform 0.3s, color 0.3s;
```

IvanFesenko (approved):
Check my comment, to avoid mistakes in the future

PR #1724, opened 2023-03-11
layout_catalog_hovers

YBM3:
- [DEMO LINK](https://ybm3.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: color, transform 300ms;
+
     &:not(:nth-child(4n)) {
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: color, transform 300ms;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
```
danielpopek94, line 27: i suggest to use cursor: pointer;

danielpopek94 (changes requested):
everything works, but for full effect I would add cursor:pointer on card;

d-wasilewski (approved):

PR #1735, opened 2023-03-14
add task solution

reseneweb:
https://reseneweb.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -21,5 +21,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+    transition: color 300ms, transform 300ms;
+
+    &:hover {
+      color: #34568b;
+      transform: scale(1.2);
+    }
   }
 }
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 300ms;
+
+    &:hover {
+      color: #00acdc;
+    }
   }
 }
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # Frontend practice with catalog page and hovers
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_catalog_hovers/)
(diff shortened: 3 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+    transition: color 300ms, transform 300ms;
```
anstsot, line 24: please add the variable for reusable styles. create separate file for it please

anstsot (changes requested):
Good job, but there is a small fix to improve your code.

witflash (changes requested):
Please add `transition: transform 300ms;` for your card because the scale is not smooth now.

image

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +19,15 @@
...
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: #{$transition-catalog};
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      color: #34568b;
```
polosanya, line 29: Create a variable for color

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 300ms;
+
+    &:hover {
+      color: #00acdc;
```
polosanya, line 44: Same here

src/styles/blocks/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$transition-catalog: color 300ms, transform 300ms;
```
polosanya, line 1: I guess there is no need to create a variable for the whole property. 
It would be better to create a variable only for transition-duration, as it is frequently used

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 300ms;
```
polosanya, line 41: Use variable for transition duration

polosanya (changes requested):
Almost perfect! Let's polish it up)

mykhalenych (approved):

PR #1740, opened 2023-03-15
Solution

superzuc:
https://superzuc.github.io/layout_catalog_hovers/


src/styles/blocks/catalog.scss
```diff
@@ -12,14 +14,21 @@
...
     box-sizing: border-box;
 
-    width: 200px;
-    height: 300px;
+    width: $card-width;
+    height: $card-height;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
polosanya, line 22: It would be great to create a variable for transition duration, as it should be the same for the whole page

src/styles/blocks/catalog.scss
```diff
@@ -12,14 +14,21 @@
...
+    transition-property: all;
```
polosanya, line 23: Don't use all. It's not good for performance. Better specify the exact properties you want add a transition to

src/styles/blocks/header.scss
```diff
@@ -38,5 +40,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
polosanya, line 43: Same here. Add correct transition-property

polosanya (changes requested):
Almost done!

src/styles/blocks/catalog.scss
```diff
@@ -12,14 +14,21 @@
...
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all $transition-duration;
+
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
```
witflash, line 29: It would be great to use `cursor: pointer;` for such cases. It's a good UX approach.

witflash (approved):
Good job! 👍

PR #1742, opened 2023-03-15
add hover transform and transition

Misha408:
- [DEMO LINK](https://misha408.github.io/layout_catalog_hovers/)


src/styles/blocks/header.scss
```diff
@@ -33,10 +33,14 @@
   }
 
   &__link {
+    transition: $transition--time;
```
vlad-demchuk, line 36: do not set transition to all effects

set for some special ones you need

vlad-demchuk: check both cases

vlad-demchuk (changes requested):

YegorKorenkov (approved):

PR #1752, opened 2023-03-17
task completed

Ozwiena:
- [DEMO LINK](https://ozwiena.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,7 +17,7 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition: transform color $transition-duration ease-out;
+    transition: all $transition-duration ease-out;
 
     &:hover {
       cursor: pointer;


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +17,14 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform color 300ms ease-out;
```
MikiOOOO, line 20: transition is not applied for cards, also consider using variable for animation time :)

MikiOOOO (changes requested):

src/styles/blocks/catalog.scss
Ozwiena: I have made a variable for transition duration, thanks for the tip :) . However, I do not understand the first part of your remark. In my code, transition is applied to every card with hover status (although it is added under the general selector, as per point 1 of the checklist). What should I change?

MikiOOOO: it is not about proper place but way of providing particular properties. (https://prnt.sc/TRQSyyF-2B0e) Try changing it to "all" and see how it works :)

MikiOOOO (changes requested):
i have left helpful comment in the last thread

src/styles/blocks/catalog.scss
Ozwiena: Now I get it, I corrected the code accordingly :)

MikiOOOO (approved):

PR #1756, opened 2023-03-18
animationsIsDone

EvgenGolub:
- [DEMO LINK](https://evgengolub.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,6 +18,27 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    @keyframes scale {
+      from {
+        transform: scale(1, 1);
+      }
+      to {
+        transform: scale(1.2, 1.2);
+      }
+    }
+
+    @keyframes color {
+      100% {
+        color: #34568b;
+      }
+    }
+
+    transform: color, scale;
+
+    &:hover {
+      animation: scale 0.3s linear, color 0.3s ease;
+    }
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,27 @@
...
+      }
+    }
+
+    @keyframes color {
+      100% {
+        color: #34568b;
+      }
+    }
+
+    transform: color, scale;
```
polosanya, line 36: Did you mean transition?

polosanya (changes requested):
Hover effects don't work correctly. Please fix it
Same with links
animation
Also, don't forget to use variables
image

src/styles/blocks/catalog.scss
```diff
@@ -1,3 +1,6 @@
+$transform-scale: 1.2;
+$card-hover-color: #34568b;
```
polosanya, line 2: Create a separate file for variables

src/styles/blocks/catalog.scss
```diff
@@ -18,6 +21,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: color 0.3s, transform 0.3s;
```
polosanya, line 24: Create a variable for transition duration, as it should be the same for the whole page and is frequently used

polosanya (changes requested):
Almost perfect!

EvgenGolub:
- [DEMO LINK](https://evgengolub.github.io/layout_catalog_hovers/)

polosanya (approved):
Well done!

PR #1763, opened 2023-03-19
task completed

diana-tuz:
[DEMO LINK](https://diana-tuz.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -15,6 +15,8 @@
 }
 
 .nav {
+  transition-duration: 300ms;
+
   &__list {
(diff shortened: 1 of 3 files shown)


src/styles/blocks/header.scss
```diff
@@ -15,6 +15,8 @@
 }
 
 .nav {
+  transition-duration: 300ms;
```
vlad-demchuk, line 18: move duration to a variable as well

vlad-demchuk (changes requested):

vlad-demchuk (approved):

PR #1769, opened 2023-03-20
finished the task

EvhenIvchenko:
- [DEMO LINK](https://EvhenIvchenko.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all 0.3s ease;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
     }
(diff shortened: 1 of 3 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: all 0.3s ease;
```
etojeDenys, line 41: try to specify transition property instead of `all`

etojeDenys (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all 0.3s ease;
```
DarkMistyRoom, line 21: list all transition properties, it is better for performance. and move 0.3s to a variable

DarkMistyRoom (changes requested):

PR #1770, opened 2023-03-21
Solution

OlhaKissa:
[DEMO LINK](https://olhakissa.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -11,6 +11,7 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
+    transition: 300ms ease;
```
IvanFesenko, line 14: By default transition is used for all. Needs to set only what changes
```suggestion
    transition: transform 0.3s ease, color  0.3s ease;
```

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 300ms ease;
```
IvanFesenko, line 41: Same here

IvanFesenko (changes requested):
Well done. Lets make some improvements

IvanFesenko (changes requested):

PR #1776, opened 2023-03-23
layout catalog hovers task implementation

Titech0:
[DEMO LINK](https://titech0.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: scale, color, $transition-time;
+    &:hover {
+      scale: 1.2;
+
+      color: $title-text-color;
+    }
+
     &:not(:nth-child(4n)) {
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: scale, color, $transition-time;
```
vlad-demchuk, line 21: it should be like that 

image

```
transition: effect duration effect duration;

File: src/styles/blocks/header.scss [hunk 2]
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: color, $transition-time;
```
vlad-demchuk, line 42: the same

vlad-demchuk (changes requested):

YegorKorenkov (approved):
Looks good

PR #1786, opened 2023-03-25
FirstCommitCatalogHovers

Kyzenko:
- [DEMO LINK](https://Kyzenko.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,10 +1,11 @@
 .catalog {
   display: flex;
   flex-wrap: wrap;
-
+  
   max-width: 944px;
   margin: 0 auto;
   padding: 50px 0;
+  perspective: 400px;
 
   &__card {
     display: flex;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +18,16 @@
...
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $duration-time, color $duration-time;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      cursor: pointer;
```
Volodymyr-Melnyk-Fron10der, line 28: just set it for default class, not :hover pseudoclass. because pointer wil be changed only on mouse hover by default

Volodymyr-Melnyk-Fron10der (changes requested):
just one remark

src/styles/blocks/catalog.scss
Kyzenko: Okey, one minutes

Kyzenko:
So yes?

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 300ms;
+  }
+
+  &__link:hover {
+    color: #00acdc;
```
OleziO, line 45: Use color variables.

OleziO (changes requested):

OleziO (approved):

PR #1790, opened 2023-03-25
add task solution

VasilPan:
- [DEMO LINK](https://VasilPan.github.io/layout_catalog_hovers/)


Volodymyr-Melnyk-Fron10der (changes requested):
set all transition properties which are animated

src/styles/main.scss
```diff
@@ -1,3 +1,4 @@
+@import "blocks/variables.scss";
```
DarkMistyRoom, line 1: better place `variables `file to `utils `folder

DarkMistyRoom (approved):
Good job

PR #1796, opened 2023-03-29
add task solution

SerhiiVoitiuk:
- [DEMO LINK](https://SerhiiVoitiuk.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -11,6 +11,7 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
+    transition: color 300ms, transform 300ms;
 
     width: 200px;
     height: 300px;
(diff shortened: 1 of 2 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -11,15 +11,22 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
+    transition: color 300ms, transform 300ms;
```
anstsot, line 14: Use SCSS variables for values, used in several places throughout the code. 
The transition time is used several times

anstsot (changes requested):
Good work, let's just create css variable

src/styles/blocks/catalog.scss
SerhiiVoitiuk: Поправив

sTorba24 (approved):
Cool!

PR #1797, opened 2023-03-29
added calalog_hovers

tatiana-sadaeva:
[DEMO LINK](https://tatiana-sadaeva.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -5,6 +5,7 @@
   max-width: 944px;
   margin: 0 auto;
   padding: 50px 0;
+  gap: 48px;
 
   &__card {
     display: flex;
@@ -16,10 +17,11 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
-
-    &:not(:nth-child(4n)) {
-      margin-right: 48px;
+   
+    &:hover {
+      transition: $duraction;
+      transform: scale(1.2);
+      color: $card-title-text-color;
     }
   }
 }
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -16,10 +17,11 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
-
-    &:not(:nth-child(4n)) {
-      margin-right: 48px;
+   
+    &:hover {
+      transition: $duraction;
```
Volodymyr-Melnyk-Fron10der, line 22: transition shoudn't be inside hover selector

set all properties which are being animated

src/styles/blocks/header.scss
```diff
@@ -8,6 +8,7 @@
   padding: 0 50px;
 
   box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+  transform: $duraction;
```
Volodymyr-Melnyk-Fron10der, line 11: don't get the purpose of it)

Volodymyr-Melnyk-Fron10der (changes requested):
hi,

1. fix your demo link, make sure you've deployed the solution correct
image

src/styles/blocks/catalog.scss
```diff
@@ -5,21 +5,23 @@
...
   margin: 0 auto;
   padding: 50px 0;
+  gap: 48px;
 
   &__card {
     display: flex;
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
+    transition: $duraction;
```
SerhiiKirik, line 15: don't use it like that, because it's the same as transition: all 0.3s
It's too hard for the browser
just list all params witch will be changed

PR #1798, opened 2023-03-31
add task solution

vovaAlieksieiev:
- [DEMO LINK](https://vovaAlieksieiev.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -11,12 +11,17 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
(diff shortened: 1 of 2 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -11,12 +11,17 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
-
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: color 300ms, transform 300ms;
```
anstsot, line 19: you're using `300ms` several times so you can create the variable for it

anstsot (changes requested):
Good work but let's create a variable

PR #1811, opened 2023-04-04
hovers

liza-strykharchuk:


Viktor-Kost (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -11,6 +11,8 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
+    transition-duration: 0.3s;
```
vlad-demchuk, line 14: would be good to move duration to variable as well

vlad-demchuk: the prev comment
image

src/styles/blocks/catalog.scss
```diff
@@ -11,6 +11,8 @@
...
+    transition-property: color, transform;
```
vlad-demchuk, line 15: use shorthand transition property

vlad-demchuk (changes requested):

IvanFesenko (approved):

PR #1812, opened 2023-04-05
solution

VadymKapeliusnyi:
- [DEMO LINK](https://VadymKapeliusnyi.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,3 +1,5 @@
+@import "./variables/variables";
+
 .catalog {
   display: flex;
   flex-wrap: wrap;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -16,10 +18,15 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    margin-bottom: $margin-catalog;
+    transition: all $transition-time $transition;
```
Viktor-Kost, line 22: it's not a good idea to use all as a value for transition-property. Specify what you'd like to change exactly

src/styles/blocks/header.scss
```diff
@@ -38,5 +40,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: $transition-time $transition;
```
Viktor-Kost, line 43: the same

Viktor-Kost (changes requested):

PR #1814, opened 2023-04-05
add task solution

Peter-Pokhylko:
- [DEMO LINK](https://Peter-Pokhylko.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -39,6 +39,7 @@
     font-size: 12px;
     font-weight: 500;
     transition-duration: 300ms;
+    transition-property: color;
 
     &:hover {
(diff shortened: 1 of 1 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
Volodymyr-Melnyk-Fron10der, line 41: set transition property as well

Volodymyr-Melnyk-Fron10der (changes requested):
just one minor fix

IvanFesenko (approved):

PR #1817, opened 2023-04-07
Develop

GlibVK:

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -22,9 +22,6 @@
 
     &:hover {
       transform: scale(1.2);
-    }
-
-    &__title:hover {
       color: #34568b;
     }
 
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: 300ms;
```
IvanFesenko, line 21: Add transition property as well. Also use shorthand for transition

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition-duration: 300ms;
+
```
IvanFesenko, line 43: Same here

IvanFesenko (changes requested):

PR #1818, opened 2023-04-07
Develop

GlibVK:

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -22,9 +22,6 @@
 
     &:hover {
       transform: scale(1.2);
-    }
-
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: 300ms;
```
anstsot, line 21: please create the css variable for transition time because you're using it few times

anstsot (changes requested):
Good work, but one of the conditions of the task was not met
_use SCSS variables for values, used in several places throughout the code_

PR #1820, opened 2023-04-07
Develop

GlibVK:

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -22,9 +22,6 @@
 
     &:hover {
       transform: scale(1.2);
-    }
-
-    &__title:hover {
       color: #34568b;
     }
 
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: $time-hover-nav-card;
```
witflash, line 21: Could you use separate `transition-property` and `transition-duration` rules here?

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: $time-hover-nav-card;
```
witflash, line 42: And here the same

witflash (changes requested):
Well done 👍 
Just let's make small improvements.

PR #1823, opened 2023-04-08
task done

zisuan-od:
https://zisuan-od.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -7,10 +7,8 @@
   padding: 50px 0;
 
   &__card {
-    display: flex;
+    @extend %display-element-properties;
     justify-content: center;
-    align-items: center;
-    box-sizing: border-box;
 
     width: 200px;
     height: 300px;
@@ -18,8 +16,16 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform $animation-duration;
+    transition-duration: $animation-duration;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+  }
+
+  &__card:hover {
+    color: #34568b;
+    transform: scale(1.2);
   }
 }
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -1,8 +1,6 @@
 .header {
-  display: flex;
-  align-items: center;
+  @extend %display-element-properties;
   justify-content: space-between;
-  box-sizing: border-box;
 
   height: 60px;
   padding: 0 50px;
@@ -38,5 +36,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $animation-duration;
+  }
+
+  &__link:hover {
+    color: #00acdc;
   }
 }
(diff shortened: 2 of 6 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -7,19 +7,25 @@
...
-    box-sizing: border-box;
 
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform $animation-duration;
+    transition-duration: $animation-duration;
```
nazarmatsevych, line 20: You already have transition-duration rule in transition rule above

src/styles/blocks/header.scss
```diff
@@ -38,5 +36,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $animation-duration;
```
nazarmatsevych, line 39: add transition rule and specify the properties you change on hover

nazarmatsevych (changes requested):

src/styles/blocks/catalog.scss
zisuan-od: Ooops. Something went wrong %)

src/styles/blocks/header.scss
zisuan-od: As I sure understand, it must be changed menu-link text-color, when it is hovered.
image
I used "color" in ":hover" and "transition" for implement transition-duration. Is it wrong?  It really works, I sweare I ckecked it :)

src/styles/blocks/catalog.scss
zisuan-od: done

src/styles/blocks/catalog.scss
```diff
@@ -7,19 +7,24 @@
...
-    align-items: center;
-    box-sizing: border-box;
 
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform $animation-duration;
```
nazarmatsevych, line 19: You change color and transform on hover so you need to specify this properties is transition rule
```suggestion
    transition: transform $animation-duration, color $animation-duration;
```

src/styles/blocks/header.scss
nazarmatsevych, line 39:
```suggestion
    transition: color $animation-duration;
```

same here

nazarmatsevych (changes requested):

src/styles/blocks/header.scss
nazarmatsevych: yeah it works, but it's better for browser to specify exactly which properties you change on hover

src/styles/blocks/catalog.scss
zisuan-od: image
Understand. When operator has two or more properties - it can set different time duration on hover
Got it, will use

src/styles/blocks/header.scss
zisuan-od: But if I set specify, when hover has only one property - it doesn't work

zisuan-od: sorry, forgot to delete "duration" from operator

zisuan-od: done.

sorry, when checked it's ability to work, forgot to delete "duration" from operator - and with specify "color" operator "animation-duration" didn't work

so I made incorrect conclusion, that in this case it will not work

vlad-demchuk (approved):

PR #1824, opened 2023-04-08
transition

shelyaa:


src/styles/blocks/catalog.scss
```diff
@@ -11,6 +13,7 @@
     justify-content: center;
     align-items: center;
     box-sizing: border-box;
+    transition: $transition-time;
```
VitaliyBondarenko1982, line 16: specify css property for transition
by default  `all` , but it is not good for performance

VitaliyBondarenko1982 (changes requested):

vlad-demchuk (approved):

PR #1825, opened 2023-04-10
add task solution

nadiiadanyleiko:
- [DEMO LINK](https://nadiiadanyleiko.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: all 0.3s ease-in-out;
+
+    &:hover {
+      color: #00acdc;
+    }
   }
 }
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # Frontend practice with catalog page and hovers
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_catalog_hovers/)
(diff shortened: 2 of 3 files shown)


src/styles/utils/__variables.scss
```diff
@@ -0,0 +1 @@
+$transition: all 0.3s ease-in-out;
```
vlad-demchuk, line 1: do not use `all`
it is better to specify the elements to which you want to add a transition

vlad-demchuk (changes requested):

src/styles/blocks/header.scss
```diff
@@ -38,5 +40,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: $transition;
```
DarkMistyRoom, line 43: set properties which transition applied for here as well

DarkMistyRoom: it is important for performance. [here is explanation](https://stackoverflow.com/questions/8947441/css3-transitions-is-transition-all-slower-than-transition-x)

DarkMistyRoom (changes requested):

src/styles/blocks/header.scss
nadiiadanyleiko: > it is important for performance. [here is explanation](https://stackoverflow.com/questions/8947441/css3-transitions-is-transition-all-slower-than-transition-x)

thanks for the explanation)

OleziO (approved):

PR #1829, opened 2023-04-11
hovers created

pr1syazhniy:
[DEMO LINK](https://pr1syazhniy.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -22,4 +22,10 @@
       margin-right: 48px;
     }
   }
+
+  &__card:hover {
+    transition-duration: 300ms;
```
Viktor-Kost, line 27: 1) move transition-duration: 300ms; to static state instead of hover one
2) add transition-property to specify what you'd like to change exactly since by default it applies to all
3) move 300ms to a variable

Viktor-Kost (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $delay;
+    transition-property: color, transform;
```
vlad-demchuk, line 21: you could use shorthand

vlad-demchuk (approved):

PR #1830, opened 2023-04-11
Add Catalog hovers

ghost:
- [DEMO LINK](https://ruslanpashkov.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: color $transition-duration;
+
+    &:hover {
+      color: #00acdc;
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


generatedman (approved):
Good job!

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: color $transition-duration;
+
+    &:hover {
+      color: #00acdc;
```
naziamoff, line 45: you can add a variable for the font as well. Yeah, you use it only once here, but it's good thing to practice creating it. Most likely that you will do it when you work on the real project

naziamoff (changes requested):
- looks wrong if I resize it down
Screenshot 2023-04-19 at 17 16 24

naziamoff (approved):

PR #1831, opened 2023-04-11
Catalog hovers

levchenko-dmytro:
- [DEMO LINK](https://levchenko-dmytro.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 300ms;
+
+    &:hover {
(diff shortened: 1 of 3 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 300ms;
```
vlad-demchuk, line 41: DRY
move transition duration to to variable

also, specify effects you need to apply transition, because it adds to all ones by default

vlad-demchuk (changes requested):

VitaliyBondarenko1982 (approved):

PR #1833, opened 2023-04-12
Develop

GlibVK:


Anastasiia-Svintsova (changes requested):
- Hover effects don't work for the navigation 😢

https://user-images.githubusercontent.com/99403198/231427827-6e732301-1c52-4c83-a00c-95093fc1a145.mov

PR #1835, opened 2023-04-12
create catalog hovers

olha-rypich:
[DEMO LINK](https://olha-rypich.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all $transition-duration;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: all $transition-duration;
```
DarkMistyRoom, line 20: please specify properties instead of using `all`

DarkMistyRoom (changes requested):
Almost there

VitaliyBondarenko1982 (approved):

PR #1838, opened 2023-04-12
add task

PashaD9:
- [DEMO LINK](https://PashaD9.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,17 @@
...
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: $duration-time;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transition: $duration-time;
```
tetiana-ivanchenko, line 27: redundant

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,17 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: $duration-time;
```
tetiana-ivanchenko, line 20: use transition property

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: $duration-time;
+
+    &:hover {
+      transition: $duration-time;
```
tetiana-ivanchenko, line 44: redundant

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: $duration-time;
```
tetiana-ivanchenko, line 41: use transition property

tetiana-ivanchenko (changes requested):

src/styles/blocks/catalog.scss
polosanya: It would be great to set the exact properties you want to apply a transition to.
Now it applies to all the properties, which is not good for the performance

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
...
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(120%);
+      cursor: pointer;
+      color: #34568b;
```
polosanya: You can set `cursor: pointer` to the default state of the element, and it will work the same.
Also, it would be great to create variables for all colours so that you can change them in one place later)

polosanya (changes requested):
Almost perfect! A few small changes left 🙂

src/styles/blocks/catalog.scss
PashaD9: It should be like this - transition: transform ease $duration-time, color ease $duration-time; ?

PashaD9: I understood it from the beginning like this - transition: all ease $duration-time;

etojeDenys (approved):
LGTM

PR #1854, opened 2023-04-19
add solution

rkhramiuk:
- [DEMO LINK](https://rkhramiuk.github.io/layout_catalog_hovers/)


.linthtmlrc.json
loralevitska, line 1: In the future - Do not add extra files to PR. To add the file to a commit use git add src/.

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
loralevitska: Combine it with `transition` and set each property for which it is used
image

loralevitska (changes requested):
Almost perfect) Let's improve

DarkMistyRoom (approved):
Well done!

PR #1856, opened 2023-04-20
add task solution

AlisaSkryp:
https://AlisaSkryp.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,3 +1,5 @@
+$duration: 300ms;
+
 .catalog {
   display: flex;
   flex-wrap: wrap;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -11,12 +10,18 @@
...
     align-items: center;
     box-sizing: border-box;
-
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transform-origin: center;
+    transition: color 300ms, transform 300ms;
```
DarkMistyRoom: move 300ms to a variable, because of using it more than one time

DarkMistyRoom (changes requested):
Just one suggestion

YegorKorenkov (approved):
Looks good

PR #1857, opened 2023-04-20
Catalog - hovers

anpiet:
[DEMO LINK](https://anpiet.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -21,5 +21,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(120%);
+      color: #34568b;
+      transition-duration: $duration;
+    }
   }
 }
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    &:hover {
+      color: #00acdc;
+      transition-duration: $duration;
+    }
   }
 }
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,4 @@
+@import "./utils/variables";
 @import "blocks/page";
 @import "blocks/header";
 @import "blocks/catalog";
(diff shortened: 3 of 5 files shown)


Dolnys (approved):
The transition works smoothly in one direction, but the card and menu hover effect doesn't return smoothly to its original size. Other than that, everything looks good.

JoyCoffeeAddict (approved):

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    &:hover {
+      color: #00acdc;
+      transition-duration: $duration;
```
JoyCoffeeAddict, line 44: The SCSS variable `$duration` is not defined in this file or in the code provided. Make sure to define this variable in a proper scope, or replace it with an actual value.

src/styles/main.scss
```diff
@@ -1,3 +1,4 @@
+@import "./utils/variables";
```
JoyCoffeeAddict, line 1: The import of the variables file is not in the wrong place. It should be placed at the top of the file, above all other imports, to ensure that the variables are accessible in all other imported files.

src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(120%);
```
JoyCoffeeAddict, line 26: The scale property should have a value of 1.2 for a 20% increase. The current value of 120% actually scales the element by 120 times.

src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,11 @@
...
+      color: #34568b;
```
JoyCoffeeAddict, line 27: The color property should not be in the hover selector of the card. It should be in the hover selector of the card title to change the card title text color to `#34568b`.

src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,11 @@
...
+      transition-duration: $duration;
```
JoyCoffeeAddict, line 28: The variable $duration is not defined in this scope. Please ensure that the variable is defined, or use a specific duration like '0.5s'.

JoyCoffeeAddict:
Don't worry about those comments, I'm just checking AI

PR #1866, opened 2023-04-24
Solution

yao157:
- [DEMO LINK](https://yao157.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -10,16 +10,24 @@
...
 
     width: 200px;
     height: 300px;
+    box-sizing: border-box;
+
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all $transition-duration;
```
vlad-demchuk, line 22: do not apply transition to all properties
specify ones you need

vlad-demchuk (changes requested):

PR #1871, opened 2023-04-26
added solution

serentinos:
- [DEMO LINK](https://serentinos.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/_variables.scss
+++ b/src/styles/_variables.scss
@@ -1,0 +1,4 @@
+$transitioin-smooth: 300ms;
+$card-hover-scale: 1.2;
+$card-accent-color: #34568b;
+$nav-link-accent-color: #00acdc;
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,7 +18,8 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition: all $transitioin-smooth ease;
+    transition: $transitioin-smooth ease;
+    transition-property: transform, color;
 
     &:hover {
       transform: scale($card-hover-scale);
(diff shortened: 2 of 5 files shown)


src/styles/_variables.scss
```diff
@@ -0,0 +1,4 @@
+$transitioin-smooth: 300ms;
+$card-hover-scale: 1.2;
+$card-accent-color: #34568b;
+$nav-link-accent-color: #00acdc;
```
panch1811, line 4: do u actually need that much of variables?
image

panch1811 (approved):
all good, check comment :)

src/styles/_variables.scss
serentinos: Probably no but i really love variables :D

src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all $transitioin-smooth ease;
```
Anna-Tkachenko: Don't use `all` in transition, it will affect in bad way on website performance
Specify only those properties that will be changed

Anna-Tkachenko (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,14 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: $transitioin-smooth ease;
+    transition-property: transform, color;
```
bohdan-mykhailenko, line 22: Interesting decision to make the code shorter, it's really hard to write transitions for few properties in one line.

bohdan-mykhailenko (approved):
Great job 👍

src/styles/blocks/catalog.scss
serentinos: @bohdan-mykhailenko The transition shorthand with multiple transition properties doesn't work in my case. That's why I just set this properties separately:)

Anna-Tkachenko (approved):
⭐️

PR #1875, opened 2023-04-26
add task solution

Bonum-taurus:
https://Bonum-taurus.github.io/layout_catalog_hovers

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,5 @@
+$duration: 300ms;
+
 @import "blocks/page";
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,5 @@
+$duration: 300ms;
```
Viktor-Kost, line 1: It is better to define the variable $duration inside the corresponding SCSS files (variables.scss)

Viktor-Kost (approved):
Good job!

PR #1884, opened 2023-04-27
Catalog hovers solution

skinnyjazz:
- [DEMO LINK](https://skinnyjazz.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -21,5 +21,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    transition: $transition;
+    &:hover{
+      transform: $card-scale;
+      color: $card-active-color;
+    }
   }
 }
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,3 +1,5 @@
+@import "./utils/variables";
+
 @import "blocks/page";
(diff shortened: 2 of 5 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,5 @@
+@import "./utils/variables";
+
```
ulositska, line 2: Good that you remembered to import variables right at the beginning:fire:

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,6 @@
+$transition: 300ms;
```
ulositska, line 1: As i understand, it's not the best practice to set transition for all, slightly better to specify exactly what will be changing.

But I think it works fine in this context. Only, in this case consider adding the 'all' keyword to your variable.

ulositska (approved):
You really did put those sass variables to a good use!

Bellow I've made one minor suggestion, but it's up to you.

src/styles/blocks/catalog.scss
```diff
@@ -21,5 +21,11 @@
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    transition: $transition;
```
silchyn, line 25: only add `transition` for needed properties

silchyn (approved):

PR #1890, opened 2023-04-27
add task solution

MaksimKad:
- [DEMO LINK](https://MaksimKad.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -2,7 +2,7 @@
   display: flex;
   flex-wrap: wrap;
 
-  max-width: 944px;
+  max-width: size($card-width, $card-margin-right);
   margin: 0 auto;
   padding: 50px 0;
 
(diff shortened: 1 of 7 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -2,7 +2,7 @@
   display: flex;
   flex-wrap: wrap;
 
-  max-width: 944px;
+  max-width: size($card-width, $card-margin-right);
   margin: 0 auto;
   padding: 50px 0;
```
julshvets: We can't set margins to block.

src/styles/blocks/catalog.scss
```diff
@@ -12,14 +12,21 @@
...
     box-sizing: border-box;
 
-    width: 200px;
+    width: $card-width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform 0.3s, color 0.3s;
```
julshvets, line 24: We can use a variable for the duration of the transform because all our elements on the page change with the same duration. But it's just a recommendation:)

src/styles/blocks/header.scss
```diff
@@ -27,9 +27,15 @@
     list-style: none;
     line-height: 60px;
 
+    transition: color 0.3s;
```
julshvets, line 30: Here you use the same duration of the transform, as in catalog.scss. We can use a variable. But it's just a recommendation.

julshvets (changes requested):

Ihor-Karpyn (approved):
FJ

PR #1891, opened 2023-04-27
add hover abd transition to card

anna-chovnova:
https://anna-chovnova.github.io/layout_catalog_hovers/


src/styles/blocks/catalog.scss
```diff
@@ -1,25 +1,34 @@
...
-
-    width: 200px;
-    height: 300px;
+    width: $card-width;
+    height: $card-height;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    margin-bottom: $card-margin;
+    transition-duration: $hover-duration;
```
etojeDenys: Add 'transition-property' to specify which properties should have the transition effect

src/styles/blocks/header.scss
```diff
@@ -26,17 +24,22 @@
...
   }
 
   &__link {
+    $hover-duration: 300ms;
     color: inherit;
     text-decoration: none;
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $hover-duration;
```
etojeDenys: Specify which properties the transition should apply to for better performance. For example, transition-property: color;

etojeDenys (changes requested):
image
the layout does not need to break on hover

src/styles/blocks/header.scss
```diff
@@ -26,17 +24,22 @@
...
   }
 
   &__link {
+    $hover-duration: 300ms;
     color: inherit;
     text-decoration: none;
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 300ms;
```
etojeDenys, line 39: Use the variable $hover-duration for the transition duration instead of the hardcoded value.

etojeDenys (approved):
nice

PR #1892, opened 2023-04-27
add task solution

k1samehoshigaki:
- [DEMO LINK](https://Mykola-Maik.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -19,7 +19,7 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition: all 0.3s linear 0s;
+    transition: transform 0.3s linear, color 0.3s linear;
 
     &:hover {
       transform: scale(120%);
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -19,11 +19,12 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition: transform 0.3s linear, color 0.3s linear;
+    transition: transform $transition-duration linear,
+      color $transition-duration linear;
 
     &:hover {
       transform: scale(120%);
-      color: $card-text-color;
+      color: $card-text-hover-color;
     }
 
     &:not(:nth-child(4n)) {
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,3 +1,4 @@
+
 .catalog {
   display: flex;
   flex-wrap: wrap;
@@ -17,6 +18,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+
+    transition: all 0.3s linear 0s;
+
+    &:hover {
+      transform: scale(120%);
+      color: $card-text-color;
+    }
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -33,7 +33,7 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
-    transition: color 0.3s linear 0s;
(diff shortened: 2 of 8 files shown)


src/styles/blocks/header.scss
```diff
@@ -34,9 +30,13 @@
 
   &__link {
     color: inherit;
-    text-decoration: none;
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 0.3s linear 0s;
```
vitvadym: `transition: color 0.3s linear;` 
i think there is no reason to specify 0s delay if it's 0)

src/styles/blocks/catalog.scss
```diff
@@ -18,6 +19,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all 0.3s linear 0s;
```
vitvadym: 👀

vitvadym (approved):
👍

src/styles/blocks/catalog.scss
silchyn: only apply `transition` to needed properties

silchyn (changes requested):
Do you really need to edit all these files?

src/styles/blocks/catalog.scss
```diff
@@ -1,3 +1,4 @@
+
```
naziamoff, line 1:
```suggestion
```

src/styles/blocks/catalog.scss
```diff
@@ -18,6 +19,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform 0.3s linear, color 0.3s linear;
```
naziamoff: let's create a variable for transition-duration, you use it a couple of times

src/styles/blocks/catalog.scss
```diff
@@ -18,6 +19,13 @@
...
+
+    &:hover {
+      transform: scale(120%);
+      color: $card-text-color;
```
naziamoff: what if this color will be needed in some different part of the project? Will this name be a good choice?

src/styles/blocks/header.scss
```diff
@@ -34,9 +30,13 @@
 
   &__link {
     color: inherit;
-    text-decoration: none;
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 0.3s linear;
```
naziamoff: transition-duration variable could be reused here as well

src/styles/blocks/header.scss
```diff
@@ -34,9 +30,13 @@
...
+
+    &:hover {
+      color: $nav-link-hover-color;
```
naziamoff: same here, what if this color will be needed in some different part of the project? Will this name be a good choice?

src/styles/main.scss
```diff
@@ -1,3 +1,7 @@
-@import "blocks/page";
-@import "blocks/header";
-@import "blocks/catalog";
+@import "./utils/reset";
+@import "./utils/normalize";
+@import "./utils/variables";
+
```
naziamoff, line 4:
```suggestion
```

naziamoff (changes requested):
not bad, but let's improve it a little

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,3 @@
+$header-nav-link-hover-color: #00acdc;
+$card-text-hover-color: #34568b;
```
naziamoff, line 2: the names are too specific, you want them more generic in your code.
like `primary-green` or sth like that. Because you can't reuse `$header-nav-link-hover-color` in some place, which is not in the header. But usually on your project you don't create a custom color for everything, you reuse existing ones

naziamoff (approved):
Good 👍  
But see my suggestion

silchyn (approved):

PR #1895, opened 2023-04-27
add task solution

SofiaLiubchenko:
- [DEMO LINK](https://SofiaLiubchenko.github.io/layout_catalog_hovers/)


PanyaPrimal (approved):
Good job with vars and transitions. 

Are json files need to be pushed here?

silchyn (approved):

silchyn (commented):
only `src/` should be included to commit

PR #1900, opened 2023-04-28
add solution

growingananas:
- [DEMO LINK](https://growingananas.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all 300ms linear;
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      transform: translateY(-20%);
+      color: $color-card-hover;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all 300ms linear;
```
silchyn: only add `transition` for needed properties

silchyn: transition duration could also be in variable

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
...
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: translateY(-20%);
```
silchyn: card should get bigger, not higher

silchyn (changes requested):

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,4 @@
+
```
naziamoff, line 1:
```suggestion
```

naziamoff (approved):
🔥

PR #1901, opened 2023-04-28
add task solution

Anastasiia971997:
- [DEMO LINK](https://Anastasiia971997.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -6,7 +6,6 @@
   margin: 0 auto;
   padding: 50px 0;
 
-
   &__card {
     display: flex;
     justify-content: center;
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -5,7 +5,7 @@
   max-width: 944px;
   margin: 0 auto;
   padding: 50px 0;
-  transition: transform $transitionDuration, color $transitionDuration;
+
 
   &__card {
     display: flex;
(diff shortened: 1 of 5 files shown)


silchyn (changes requested):
Don't forget to commit and push your solution! Currently, only `readme.md` is changed

KatyaZaremska (approved):

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition: color $transitionDuration;
+
+    &:hover {
+      color: $navigationlinkTextColor;
```
naziamoff: what if you will need to reuse this color later in some other part of the project? Will this name be good enough for that purpose?

naziamoff (changes requested):
Actually the code looks not bad in general, but there are definitely things to fix

- transitioning looks very wrong

https://user-images.githubusercontent.com/103033758/235613742-f535f691-86e1-4c8a-b531-701fd573b2b8.mov

src/styles/blocks/catalog.scss
```diff
@@ -6,6 +6,7 @@
   margin: 0 auto;
   padding: 50px 0;
 
+
```
Anastasiia971997: I saw an extra indent, as I already added

silchyn (approved):

PR #1902, opened 2023-04-28
add task solution

MaksymBevskyi:
[DEMO ](https://MaksymBevskyi.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -10,16 +10,23 @@
...
     align-items: center;
-    box-sizing: border-box;
 
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform $transform-duration;
```
polosanya: What about color? It should change smoothly too

src/styles/blocks/header.scss
```diff
@@ -38,5 +37,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition-duration: $transform-duration;
```
polosanya:
```suggestion
    transition: color $transform-duration;
```

polosanya (changes requested):
Almost perfect!

Volodymyr-Melnyk-Fron10der (approved):

PR #1905, opened 2023-04-28
made layout catalog hovers

vladyslav-lapshyn:
- [DEMO LINK](https://vladyslav-lapshyn.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/utils/_variables.scss
+++ b/src/styles/utils/_variables.scss
@@ -1,1 +1,5 @@
 $main-transition-duration: 300ms;
+
+$color-main-accent: #060b35;
+$color-hover-link: #00acdc;
+$color-hover-card: #34568b;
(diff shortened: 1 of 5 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$hover-duration: 300ms;
```
ghost: it would be better to say that it's exactly the transition duration

ghost (approved):

src/styles/utils/_variables.scss
vladyslav-lapshyn: @ruslanpashkov that is a good recommendation, thanks

Ihor-Karpyn (approved):
GJ, let's create variavles for colors

PR #1909, opened 2023-04-29
complete a task

Duster-1:
https://Duster-1.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,7 +18,7 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition-duration: $transition-duration;
+    transition: color $transition-duration, transform $transition-duration;
 
     &:hover {
       color: #34568b;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $transition-duration;
```
nazarmatsevych: You also need to specify the properties you change on hover in transition rule.

Example:

`transition: color $transition-duration, transform $transition-duration;`

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $transition-duration;
```
nazarmatsevych, line 41: same here

nazarmatsevych (changes requested):

VitaliyBondarenko1982 (approved):

PR #1911, opened 2023-04-30
add task solution

LeonIllya:
- [DEMO LINK](https://LeonIllya.github.io/layout_catalog_hovers/)


amashutikov (changes requested):
Good job. But your demo-link is not working(

Ihor-Karpyn (changes requested):
- do we need to change card backgrount on hover?
image

- animation should be smooth

>- Smoothly (during 300ms) increase the card by 20 percent on hover (neighboring cards should not be affected)
>- Smoothly (during 300ms) change card title text color to #34568b


 
 


- create variables for colors

Ihor-Karpyn (commented):
fix demolink pls

PR #1912, opened 2023-04-30
(ym) catalog hovers

YeMyr:
- [DEMO LINK](https://yemyr.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,8 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition: color $trans-duration $smooth,
-      transform $trans-duration $smooth;
+    transition: color $transition-duration $transition-timing-function,
+      transform $transition-duration $transition-timing-function;
 
     &:not(:nth-child(4n)) {
       margin-right: $space-btwn-cards;
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -2,9 +2,9 @@
   display: flex;
   flex-wrap: wrap;
 
-  max-width: 944px;
+  max-width: $catalog-width;
   margin: 0 auto;
-  padding: 50px 0;
+  padding: $catalog-padding-y 0;
 
   &__card {
     display: flex;
@@ -12,14 +12,22 @@
     align-items: center;
     box-sizing: border-box;
 
-    width: 200px;
+    width: $card-width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $trans-dur;
+    transition-timing-function: $smooth;
+
     &:not(:nth-child(4n)) {
-      margin-right: 48px;
+      margin-right: $card-space;
+    }
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
     }
   }
 }
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,11 +18,11 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition-duration: $trans-dur;
-    transition-timing-function: $smooth;
+    transition: color $trans-duration $smooth,
+      transform $trans-duration $smooth;
 
     &:not(:nth-child(4n)) {
-      margin-right: $card-space;
+      margin-right: $space-btwn-cards;
     }
 
     &:hover {
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -23,7 +23,7 @@
 
   &__item {
     position: relative;
-    color: #060b35;
+    color: $color-main-accent;
     list-style: none;
     line-height: 60px;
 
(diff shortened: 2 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -2,24 +2,32 @@
...
 
-    width: 200px;
+    width: $card-width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $trans-dur;
+    transition-timing-function: $smooth;
```
ghost: in general, it is good to use the properties individually, but you can also write in shortened form

ghost (changes requested):
try to change your `font-size` or any other properties, your transition will also affect it, let's fix it

https://user-images.githubusercontent.com/129327901/235376557-a51a68e4-0744-44b3-a34e-5adc6af100d5.mov

src/styles/blocks/catalog.scss
YeMyr: Sorry, I did not get what you are showing me by turning off the margins and how do fonts relate to this...
Can you explain more in detail what you mean?

as for the short form, it is not as readable as individual properties.

ghost: your transition is also applied to your margins, but it should affect only your card size and text colors

such transitions may appear at unnecessary moments (as in the example with margins) and degrade performance due to permanent calculations of these properties

ghost: try to specify which properties should use transitions with `transition-property`

KatyaZaremska (approved):

src/styles/blocks/catalog.scss
```diff
@@ -2,24 +2,32 @@
...
     box-sizing: border-box;
 
-    width: 200px;
+    width: $card-width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $trans-dur;
```
naziamoff: it is a good practice to avoid such a shortcuts. You want to keep your code maximally clear
```suggestion
    transition-duration: $trans-duration;
```

src/styles/blocks/catalog.scss
naziamoff: actually @ruslanpashkov is right, you should specify exact properties you want to transition

src/styles/blocks/catalog.scss
```diff
@@ -2,24 +2,32 @@
...
+    transition-timing-function: $smooth;
+
     &:not(:nth-child(4n)) {
-      margin-right: 48px;
+      margin-right: $card-space;
```
naziamoff: what's card-space? not clear to me what does this variable mean - try sth more descriptive

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,12 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition-duration: $trans-dur;
+    transition-timing-function: $smooth;
+
```
naziamoff: let's combine everything transition-related in this block into a single property

naziamoff (changes requested):
not bad, but requires some polishing 💅

src/styles/blocks/header.scss
YeMyr: fixed

src/styles/blocks/catalog.scss
YeMyr: fixed

src/styles/blocks/catalog.scss
YeMyr: fixed

src/styles/blocks/catalog.scss
```diff
@@ -2,24 +2,32 @@
...
 
-    width: 200px;
+    width: $card-width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: color $trans-duration $smooth,
+      transform $trans-duration $smooth;
```
naziamoff: my bad, I did not correct this name fully during the last review. When I speak of maximally clear variable names I mean that:
```suggestion
    transition: color $transition-duration $transition-timing-function,
      transform $transition-duration $transition-timing-function;
```

src/styles/blocks/catalog.scss
```diff
@@ -2,24 +2,32 @@
...
+
     &:not(:nth-child(4n)) {
-      margin-right: 48px;
+      margin-right: $space-btwn-cards;
```
naziamoff, line 25: not the best naming option either. I would say you don't need this at all, we usually don't create variables for margins. There's atomic CSS libs for that, so you could just add classname `mr-48` and it will apply the needed margin. But for now I would say it is better to stick to usual margin property

naziamoff (approved):
good job, actually. But see some suggestions and consider them in the future

src/styles/blocks/catalog.scss
YeMyr: ok, I'll consider it in the future.
As for now, I can't drop it to usual margin property, as this variable is used in the catalog width calculation.
image

src/styles/blocks/catalog.scss
YeMyr: Ok, thanks, Ive got it.
This one is fixed.

PR #1914, opened 2023-04-30
add task solution

kirilshorin:
- [DEMO LINK](https://kirilshorin.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,3 +1,5 @@
+$hover-color: #34568b;
+
 .catalog {
   display: flex;
   flex-wrap: wrap;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +20,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform 300ms, color 300ms;
```
naziamoff, line 23: would be a nice idea to create a variable for transition-duration as well, as usually you use same transition-duration across all your project

src/styles/blocks/catalog.scss
```diff
@@ -1,3 +1,5 @@
+$hover-color: #34568b;
```
naziamoff, line 1: usually we give more generic names for our css-variables. If you gonna need to use this color not on hover - will the name be a good choice?

src/styles/blocks/header.scss
```diff
@@ -1,3 +1,5 @@
+$link-color: #00acdc;
```
naziamoff, line 1: same here regarding the name for a variable. It is usually `$primary-green`, `$secondary-green`, or sth like that

naziamoff (approved):
good job! See minor suggestions

PR #1916, opened 2023-05-01
add task solution

dimoshen:
- [DEMO LINK](https://dimoshen.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,11 +17,11 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition: transform 300ms, color 300ms;
+    transition: transform $card-element-transition, color $card-element-transition;
 
     &:hover {
-    transform: scale(1.2);
-    color: #34568b;
+      transform: scale(1.2);
+      color: #34568b;
     }
 
     &:not(:nth-child(4n)) {
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms, color 300ms;
+
+    &:hover {
+    transform: scale(1.2);
+    color: #34568b;
+    }
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,11 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition: transform $card-element-transition, color $card-element-transition;
+    transition: transform $card-element-transition,
+      color $card-element-transition;
 
     &:hover {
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms, color 300ms;
```
ghost: checklist p4

> [SASS] - use variables for the main values so that you'll be able to reuse them and give them descriptive names. But don't overuse them, don't create variable for the value that's used just once.

ghost (changes requested):

src/styles/blocks/catalog.scss
dimoshen: added transition variable

src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $card-element-transition, color $card-element-transition;
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
```
naziamoff: would be nice to create a variable for this color

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 300ms;
+
+    &:hover {
+      color: #00acdc;
```
naziamoff: and for this one as well

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: color 300ms;
```
naziamoff: variable could be created for the transition-duration as well

naziamoff (changes requested):
not bad 👍 
But demo is not working

src/styles/blocks/header.scss
dimoshen: created variables for header and card transitions

src/styles/blocks/header.scss
dimoshen: variable created

src/styles/blocks/catalog.scss
dimoshen: variable created

dimoshen:
> I've fixed the demo link. Here's a copy:

- [DEMO LINK](https://dimoshen.github.io/layout_catalog_hovers/)

src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $card-element-transition,
```
naziamoff, line 20: usually the same transition duration is used across the whole project, so there's no need to specify the place where you use it. As variables are created for reusability - you should give them more generic names
```suggestion
    transition: transform $transition-duration,
```

naziamoff (approved):
🔥

PR #1917, opened 2023-05-01
add task solution

liza330:
[DEMO LINK](https://liza330.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,7 +17,7 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition-duration: $transition-time;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $transition-time;
```
YegorKorenkov: It's more correct to specify for which rule the transition time is set, for example: `transition: color $transition-time`

YegorKorenkov (changes requested):
Please fix minor mistake

YegorKorenkov (approved):
Looks good

PR #1920, opened 2023-05-01
Task solved

olehmahera:
- [DEMO LINK](https://olehmahera.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform $transDur, color $transDur ;
+
+    &:hover {
+      transform: scale(1.2);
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: transform $transDur, color $transDur ;
```
naziamoff, line 21: you want your code to be maximally clear for those who read it
```suggestion
    transition: transform $transition-duration, color $transition-duration ;
```

naziamoff: but good idea to create a variable 👍

naziamoff (approved):
well done 👍 

Also, would be nice to add variables for colors

PR #1922, opened 2023-05-01
add task solution

KatyaZaremska:
- [DEMO LINK](https://KatyaZaremska.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -16,10 +16,17 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    margin-bottom: $cardMargin;
+
+    transition: color $transitionDuration, transform $transitionDuration;
+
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -16,10 +16,17 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    margin-bottom: $cardMargin;
```
naziamoff, line 19: it is a great idea to create variables, but in my opinion variables for margins like that are redundant

src/styles/blocks/catalog.scss
```diff
@@ -16,10 +16,17 @@
...
+
+    transition: color $transitionDuration, transform $transitionDuration;
```
naziamoff, line 21: 🔥

naziamoff (approved):
looks good, but see comments below

PR #1923, opened 2023-05-01
added task solution catalog hovers

holytimy4:
[DEMO LINK](https://holytimy4.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/variables.scss
+++ b/src/styles/variables.scss
@@ -1,0 +1,3 @@
+$cardTitleColor: #34568b;
+$transitionDuration: 0.3s;
+$navigationLinkColor: #00acdc;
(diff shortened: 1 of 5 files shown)


src/styles/variables.scss
```diff
@@ -0,0 +1,3 @@
+$cardTitleColor: #34568b;
+$transitionDuration: 0.3s;
+$navigationLinkColor: #00acdc;
```
naziamoff, line 3: we usually use more generic names for color variables. If you will need the same color at some point, but not in the context of the card title color - will this name be a good choice? That's why we usually have it like `$primary-green`, `$secondary-green`, etc.

naziamoff (approved):
Good job! See one suggestion:

PR #1925, opened 2023-05-02
catalog_hover

Krivets:


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: $hover-timing;
```
vlad-demchuk: specify properties you need to apply transition to

vlad-demchuk (changes requested):

src/styles/blocks/catalog.scss
tetiana-ivanchenko: use transition shorthand

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,12 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+
+    transition-property: color;
+    transition-duration: $hover-timing;
```
tetiana-ivanchenko: use transition shorthand

tetiana-ivanchenko (changes requested):

OleziO (approved):

PR #1928, opened 2023-05-02
added hovers to header links and cards

vladyslav10shevchenko:
- [DEMO LINK](https://sheva10barca.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/utils/_variables.scss
+++ b/src/styles/utils/_variables.scss
@@ -1,4 +1,3 @@
-$transitionTimeHeader: 300ms;
-$transitionTimeCard: 300ms;
+$transitionDuration: 300ms;
 $colorLinkHeader: #00acdc;
 $colorTitleCard: #34568b;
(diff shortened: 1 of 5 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,4 @@
+$transitionTimeHeader: 300ms;
+$transitionTimeCard: 300ms;
```
Anastasiia-Svintsova: Why do you need 2 separate variables for the same value? 
```suggestion
$transition-duration: 300ms;
```

Anastasiia-Svintsova (changes requested):
Great job! Let's add some improvements

src/styles/utils/_variables.scss
vladyslav10shevchenko: in case we wanna have different durations in the future

nazarmatsevych (approved):

PR #1934, opened 2023-05-03
commit

bohdanmayor:
- [DEMO LINK](https://bohdanmayor.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -23,9 +23,7 @@
       margin-right: 48px;
     }
 
-    transition-property: transform;
-    transition-duration: 300ms;
-    transition-timing-function: linear;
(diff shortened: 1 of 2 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -21,5 +22,14 @@
...
       margin-right: 48px;
     }
+
+    transition-property: transform;
+    transition-duration: 300ms;
+    transition-timing-function: linear;
+  }
+
+  &__card:hover {
+    color: #34568b;
```
IvanFesenko: You also changing color< but trasition is only for trasform. Also you can use short hand
`transition: color 0.3s, transform 0.3s linear`

IvanFesenko (changes requested):
Good job, small improvement required

mykhalenych (approved):

PR #1938, opened 2023-05-04
catalog with hover get done

LukashevychSergey:
- [DEMO LINK](https://Lukashevychsergey.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,6 +17,14 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition:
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,14 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition:
+    color $transition-time ease,
+    transform  $transition-time ease;
```
IvanFesenko, line 22: will be same
```suggestion
    transition: color $transition-time, transform $transition-time ease;
```

IvanFesenko (approved):

PR #1939, opened 2023-05-04
add task catalog_hovers solution

yuliiabinkovska:
- [DEMO LINK](https://yuliiabinkovska.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,3 +1,5 @@
+$hover-card-title-color: #34568b;
+
 .catalog {
   display: flex;
   flex-wrap: wrap;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -1,3 +1,5 @@
+$hover-card-title-color: #34568b;
```
polosanya: Let's create a separate file to store variables

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +20,16 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: 300ms;
```
polosanya: It would be great to create a variable for transition duration as it should be the same for whole page

polosanya (changes requested):
Almost done!

VitaliyBondarenko1982 (approved):

PR #1940, opened 2023-05-05
add smooth hover

a-kovalov:
- [DEMO LINK](https://a-kovalov.github.io/layout_catalog_hovers/)

NB: 
1. I decided to not add `transform-origin: center` because it's a default value and matches with requirements of the task;
2. In my opinion there is no sense to add a `transition-delay` property because it will be too slow.
3. I used `linear` - because as I understand the task to make all changes smooth it makes with the same speed from start to end changes.


etojeDenys (approved):
lgtm

PR #1941, opened 2023-05-05
Develop

valerkameladze:
- [DEMO LINK](https://valerkameladze.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms;
```
anstsot: please create variables for repeatable values (300ms)

anstsot (changes requested):
It’s almost done! Let’s improve a little

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +19,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $smooth;
```
IvanFesenko, line 22: As you change color on hover too, so needs to add color to transition as well
```suggestion
    transition: transform $smooth, color $smooth;
```

IvanFesenko (approved):

PR #1948, opened 2023-05-09
add task solution1

FIREDAN99:
https://FIREDAN99.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -1,3 +1,5 @@
+@import "../utils/variables";
+
 .header {
   display: flex;
   align-items: center;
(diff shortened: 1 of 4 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +40,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: 300ms;
```
etojeDenys: The transition property should include what CSS properties it should apply to. For example, it could be 'transition: color 300ms;' to make it clear that the transition applies to the color property.

etojeDenys (changes requested):

mykhalenych (changes requested):
Did you push your code?
I don't see any changes from previous review

sTorba24 (approved):
GJ 🚀

PR #1952, opened 2023-05-10
layout_catalog_hovers

aamelnyk:
- [DEMO LINK](https://aamelnyk.github.io/layout_catalog_hovers/)


src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition: $transition;
```
etojeDenys: Add a specific transition property, duration, and easing function to the transition shorthand property for better control over the transition effect.

etojeDenys (changes requested):

mykhalenych (approved):

PR #1957, opened 2023-05-11
Add task solution

kbekher:
- [DEMO LINK](https://kbekher.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: $transition-body;
```
IvanFesenko: By default transition is applienig to all - thats is hard for browser. 
Your should set only values you change.
Fix same a bit bottom
```suggestion
    transition: color $transition-duration, transform $transition-duration;
```

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$transition-body: all 300ms;
```
IvanFesenko:
```suggestion
$transition-duration: 300ms;
```

IvanFesenko (changes requested):

IvanFesenko (approved):

PR #1958, opened 2023-05-11
add transform

Leva313:
- [DEMO LINK](https://Leva313.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -5,6 +5,7 @@
   max-width: 944px;
   margin: 0 auto;
   padding: 50px 0;
+  perspective: 400px;
 
   &__card {
     display: flex;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +18,13 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms;
+    transition-duration: 300ms;
```
IvanFesenko: You already set duration, also you change color< so please add it.
make same for header
```suggestion
    transition: transform 0.3s, color 0.3s;
```

IvanFesenko (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,6 +18,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 0.3s, color 0.3s;
```
Viktor-Kost, line 21: move `0.3s` to a variable since you've used this value three times

Viktor-Kost (approved):

PR #1959, opened 2023-05-11
add task solution

h1deeeee:
- [DEMO LINK](https://h1deeeee.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -5,6 +5,7 @@
   max-width: 944px;
   margin: 0 auto;
   padding: 50px 0;
+  perspective: 100px;
 
   &__card {
     display: flex;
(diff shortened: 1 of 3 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,6 +18,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms, color 300ms;
+
+    &:hover {
+      color: #34568b;
+      transform: translateZ(20px);
```
witflash: It's better to use `transform: scale(1.2);`. TranslateZ is not what you need.

witflash (changes requested):
Let's improve your solution a bit.

Use `scale` on hover. It looks weird for now:
image

src/styles/blocks/catalog.scss
```diff
@@ -17,6 +17,12 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms, color 300ms;
```
YegorKorenkov, line 20: More correctly move repeated values to variables

YegorKorenkov (approved):
Nice work!

PR #1960, opened 2023-05-12
add task layout_catalog_hover

volodymyrsar:
- [DEMO LINK](https://volodymyrsar.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,8 +17,7 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition-duration: 300ms;
-    transition-property: transform, color;
+    transition: transform 300ms, color 300ms;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,18 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,18 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: 300ms;
```
VitaliyBondarenko1982: move transition-durarion to variable, as this value uses two times

VitaliyBondarenko1982 (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,17 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms, color 300ms;
```
sTorba24: mentor meant to create a variable for 300ms, because it's used 3 times so would be good to not repeat

sTorba24 (changes requested):
GJ 🚀 Little comment to fix

src/styles/main.scss
```diff
@@ -1,3 +1,4 @@
 @import "blocks/page";
 @import "blocks/header";
 @import "blocks/catalog";
+@import "./_variables"
```
IvanFesenko: To avoid errors variables should be first
```suggestion
@import "./_variables"
@import "blocks/page";
@import "blocks/header";
@import "blocks/catalog";
```

IvanFesenko (approved):

PR #1963, opened 2023-05-12
add task solution

ilona-borkovska:
- [DEMO LINK](https://ilona-borkovska.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,7 +18,8 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-    transition: $transition-time ease;
+    transition: transform $transition-time ease,
+      color $transition-time ease;
 
     &:not(:nth-child(4n)) {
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -11,15 +11,22 @@
...
     align-items: center;
     box-sizing: border-box;
+    position: relative;
 
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: $transition-time ease;
```
YegorKorenkov: it is more correct to enumerate the properties for which it will be applied transition:
`transition: transform $transition-time ease, color $transition-time ease`

YegorKorenkov (approved):
Nice work, please pay attention on remark

src/styles/blocks/catalog.scss
ilona-borkovska: I've fixed it, thanks)

PR #1972, opened 2023-05-15
catalog hovers

antonshtef:
- [DEMO LINK](https://antonshtef.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,7 +18,7 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
-    transition: all $transition-time;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: all $transition-time;
```
IvanFesenko: Do not set all propertie, set only those values you will change
```suggestion
    transition: transform $transition-time, color $transition-time;
```

IvanFesenko (changes requested):
Fix small remark

mykhalenych (approved):

PR #1973, opened 2023-05-15
transition and transform were added

Yuliia-Hrinchenko1:
- [DEMO LINK](https://Yuliia-Hrinchenko1.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $transition-duration;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      transform: scale(120%);
+      color: #34568b;
     }
   }
 }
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $transition-duration;
```
nazarmatsevych:
```suggestion
    transition-duration: transform $transition-duration, color $transition-duration;
```

You also need to specify the properties you change on hover

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: $transition-duration;
```
nazarmatsevych: same here

nazarmatsevych (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
...
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $transition-duration, color $transition-duration;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(120%);
```
volodymyr-soltys97, line 27: The scale value should be 1.2 instead of 120% to increase the card size by 20% on hover.

volodymyr-soltys97 (approved):
Good job

PR #1975, opened 2023-05-15
create animations

ivan-baranovskyi:
- [DEMO LINK](https://ivan-baranovskyi.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -9,6 +9,8 @@
 
   box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
 
+
+
   &__logo {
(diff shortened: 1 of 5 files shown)


src/styles/blocks/header.scss
```diff
@@ -21,15 +23,22 @@
...
+    transition: color $animationDuration;
+
+    &:hover {
+      color: #00acdc;
+    }
 
     &:not(:last-child) {
       margin-right: 20px;
     }
+
```
TarasHoliuk, line 41: this line is redundant
also, only one line should be between selectors or selector's rules
FYI

TarasHoliuk (approved):

PR #1976, opened 2023-05-16
add task solution

monikaprochniewicz:
- [DEMO LINK](https://monikaprochniewicz.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -18,8 +18,17 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: $transition-setting;
+  
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
   }
+
+  &__card:hover {
+    transform: scale(1.2);
+    color: #34568b;
+  }
+
 }
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -1,3 +1,5 @@
+@import "../utils/variables";
+
 .header {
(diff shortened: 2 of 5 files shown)


src/styles/blocks/header.scss
```diff
@@ -38,5 +40,11 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    
+    transition: $transition-setting;
```
DorotaLeniecDev, line 44: This is not working, as transition only is set to work for `transform` property. You can change it and set to `all` so it works for all the properties that are changing

src/styles/utils/variables.scss
```diff
@@ -0,0 +1 @@
+$transition-setting: transform 300ms linear;
```
DorotaLeniecDev:
```suggestion
$transition-setting: all 300ms linear;
```

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,17 @@
...
+  
+
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
   }
+
+  &__card:hover {
+    transform: scale(1.2);
+    color: #34568b;
```
DorotaLeniecDev, line 31: This transition is also not happening smoothly due to the fack I described in prev comment

DorotaLeniecDev (changes requested):
Changes needed

Radoslaw-Czerniawski (approved):
Works like it should :)

PR #1981, opened 2023-05-18
Develop

Klishchar-Ihor:
- [DEMO LINK](https://Klishchar-Ihor.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $transition-duration-value;
+    transition-property: transform, color;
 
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform $transition-duration-value;
+    transition-property: transform, color;
```
vlad-demchuk, line 21: you can use transition in this way
image

vlad-demchuk: https://www.w3schools.com/css/css3_transitions.asp

vlad-demchuk (changes requested):

PR #1982, opened 2023-05-18
catalog_hovers

KarolinaWiktoria:
https://KarolinaWiktoria.github.io/layout_catalog_hovers/

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,15 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition-duration: $transition-time;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      transform: scale(120%);
+      color: #34568b;
     }
   }
 }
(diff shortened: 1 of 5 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,15 @@
...
     margin-bottom: 48px;
+    transition-duration: $transition-time;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(120%);
+      color: #34568b;
```
Radoslaw-Czerniawski: All the colors could be variables

src/styles/main.scss
```diff
@@ -1,3 +1,5 @@
+@import "./utils/variables.scss";
+
```
Radoslaw-Czerniawski: There is no need for an empty line here

Radoslaw-Czerniawski (approved):
Very good, left few comments, keep in mind for future reference

src/styles/main.scss
KarolinaWiktoria: already fixed :). Thank you!

PR #1993, opened 2023-05-29
add task solution

spojrzenie:


src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition: 0.3s ease;
```
DorotaLeniecDev: Specify what will be transitioned, you can use all, or deal with transition for color and transform separately

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,3 @@
+$nav-link-hover: #00acdc;
+$card-title-hover: #34568b;
+$card-increase-hover: scale(120%);
```
DorotaLeniecDev: Do not overuse variables. If value is used once then there is no need to make variable out of it. Also name variables so they are more descriptive for example:
```suggestion
$color-blue-light: #00acdc;
$color-blue: #34568b;
```

DorotaLeniecDev (changes requested):
few changes needed

src/styles/blocks/catalog.scss
```diff
@@ -18,8 +18,15 @@
     border-radius: 5px;
     margin-bottom: 48px;
```
darokrk, line 20: remove this blank white space, it's redundant here

darokrk (approved):
Nicely done 👍🏽 ! Added one mention but it's not necessary to fix :)

PR #1997, opened 2023-06-05
Catalog-hovers

Kacper-Leman:
https://Kacper-Leman.github.io/layout_catalog_hovers/  -Demo Link

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -39,4 +39,10 @@
     font-size: 12px;
     font-weight: 500;
   }
+
+  &__link:hover {
+    transition: color;
+    transition-duration: $delayValue;
+    color: #00acdc;
+  }
 }
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -41,8 +41,7 @@
   }
 
   &__link:hover {
-    transition: color;
-    transition-duration: $delayValue;
+    transition: color $delayValue;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/header.scss
```diff
@@ -39,4 +39,10 @@
     font-size: 12px;
     font-weight: 500;
   }
+
+  &__link:hover {
+    transition: color;
+    transition-duration: $delayValue;
```
DorotaLeniecDev: Add transition to origin element (not :hover) so the transisiton is smooth in both directions, you can use  short version:
```suggestion
    transition: color $delayValue
```

DorotaLeniecDev (changes requested):
One small change needed

src/styles/blocks/header.scss
```diff
@@ -39,4 +39,9 @@
     font-size: 12px;
     font-weight: 500;
   }
+
+  &__link:hover {
+    transition: color $delayValue;
+    color: #00acdc;
+  }
```
DorotaLeniecDev: Put transition to origin element not hover to make animation smooth in both directions:
```suggestion
  &__link {
    transition: color $delayValue;  
    }
```

DorotaLeniecDev (changes requested):
One more fix needed

darokrk (approved):
Approved 👍🏽  good job 🥇

PR #2005, opened 2023-06-28
added solution

AnastasiiaCagle:
[DEMO LINK](https://AnastasiiaCagle.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -17,9 +17,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms, color 300ms;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
+    }
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
+      cursor: pointer;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -17,9 +17,16 @@
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
+    transition: transform 300ms, color 300ms;
```
Anastasiia-Svintsova: It's better to move the transition duration to a variable, you use it not once.

Anastasiia-Svintsova (changes requested):
Almost done! Let's add an improvement

src/styles/blocks/catalog.scss
```diff
@@ -17,9 +19,16 @@
...
     margin-bottom: 48px;
+    transition: transform $duration, color $duration;
 
     &:not(:nth-child(4n)) {
       margin-right: 48px;
     }
+
+    &:hover {
+      transform: scale(1.2);
+      color: #34568b;
```
polosanya, line 30: It would be great to create variables for colors too

polosanya (approved):
Good job!

PR #2009, opened 2023-07-14
Catalog Hover Solution

aRabiej:
- [DEMO LINK](https://aRabiej.github.io/layout_catalog_hovers/)

### DIFF
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -21,7 +21,7 @@
   font-size: 12px;
   font-weight: 500;
 
-  transition: color, $transitionDuration;
+  transition: color $transitionDuration;
 
   &:hover {
     color: $color-link-hover;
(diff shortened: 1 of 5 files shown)


src/styles/variables.scss
```diff
@@ -0,0 +1,3 @@
+$cardHeadercolor: #34568b;
```
DorotaLeniecDev: Use standard name for naming color variable like ```color-blue```

src/styles/blocks/header.scss
```diff
@@ -1,42 +1,29 @@
...
+  text-decoration: none;
+  text-transform: uppercase;
+  font-size: 12px;
+  font-weight: 500;
 
-    &:not(:last-child) {
-      margin-right: 20px;
-    }
-  }
+  transition: color, $transitionDuration;
```
DorotaLeniecDev: This is incorrect, comma should separate the transitions group
```suggestion
  transition: color $transitionDuration;
```

DorotaLeniecDev (changes requested):
Add demo link

DorotaLeniecDev (changes requested):
Demo page is still not working :(

PR #2021, opened 2023-10-04
catalog_hover

usernameluke:
- [DEMO LINK](https://usernameluke.github.io/layout_catalog_hovers/)


src/styles/blocks/catalog.scss
```diff
@@ -18,6 +18,13 @@
     border-radius: 5px;
     margin-bottom: 48px;
 
+    transition-duration: 300ms;
```
anastasiia-tilikina: it's better to set transitioned properties explictly

src/styles/blocks/header.scss
```diff
@@ -38,5 +38,10 @@
     text-transform: uppercase;
     font-size: 12px;
     font-weight: 500;
+    transition-duration: 300ms;
```
anastasiia-tilikina: fix here as well

anastasiia-tilikina (changes requested):
Almost done!
let's just add some improvements

src/styles/blocks/catalog.scss
lerastarynets: not fixed

src/styles/blocks/header.scss
lerastarynets: not fixed

lerastarynets (changes requested):
Pls fix prev comments

anastasiia-tilikina (approved):
GJ!

